PR #2, opened 2015-06-01
Fixed async issue using promises

psayre23:
This will work in node v0.12 and on. Will need a library for backwards compatibilty.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,16 @@
 var alexa={};
 
 alexa.response = function() {
-	var prom = null, resolve, reject;
+	var isWaiting = false;
+	var resolve, reject;
+	var prom = new Promise(function (_resolve, _reject) {
+		resolve = _resolve;
+		reject = _reject;
+	});
+	prom.then(function () {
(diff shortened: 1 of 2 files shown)


jaredalexander:
Thanks for the quick fix. How can I utilize this? Inside my app.intent function, I have an async http request that talks to an API that returns JSON. On success, I parse the JSON and grab the result. I then call say("blah") and shouldEndSession(true), and lastly done(). However, node then crashes when I call done(). I get the following error:

TypeError: Cannot read property 'then' of null
(line 150 of your app.js)

Any help would be greatly appreciated, thanks!

psayre23:
Ah! The problem is the promise doesn't exist yet. I'll take a quick look at
a fix.

~ Paul

psayre23:
That should fix the issue. I was only testing it async, so I never saw it. Good catch.

matt-kruse:
Thanks for the patch. I resolved this issue using Promises, but in a slightly different way. Update coming soon.

PR #23, opened 2015-09-26
added exhaustiveUtterances option

jpriebe:
I love alexa-app's ability to generate utterances.  It has helped me build very large lists of utterances that would be impossible to manage by hand.

I have run into limitations, though, when I have utterances with a number of shortcuts plus a number of slots with their own dictionaries.  The cartesian product gets so large that I can't post all the utterances in my Alexa Skill's interaction model.

Some of this stuff is not well documented, but I have been doing a lot of reading, and other developers have indicated that you don't need to include every sample value in every variant of the utterance.  So we don't need a strict cartesian product.

I have added a configuration option that lets you control whether a full cartesian product is generated.  The default is to *not* generate the full cartesian product.  It generates the full cartesian product of the *shortcuts*, and then fills in sample values for the slots.

Let's look at a quick example.

```
app.dictionary = {
    "movie_names": ["star wars", "inception", "gattaca", "the matrix"]
};

app.intent('DemoIntent',
    {
        "slots": {
            "MOVIE":"LITERAL"
        },
        "utterances": [
            "{foo|bar|baz} {foo|bar|baz} {movie_names|MOVIE}"
        ]
    },
    function (request, response) {
    }
);
```

Here are the default utterances generated by my modified code:

```
DemoIntent	foo foo {star wars|MOVIE}
DemoIntent	bar foo {inception|MOVIE}
DemoIntent	baz foo {gattaca|MOVIE}
DemoIntent	foo bar {the matrix|MOVIE}
DemoIntent	bar bar {star wars|MOVIE}
DemoIntent	baz bar {inception|MOVIE}
DemoIntent	foo baz {gattaca|MOVIE}
DemoIntent	bar baz {the matrix|MOVIE}
DemoIntent	baz baz {star wars|MOVIE}
```

Note that every combination of the two shortcuts are generated (3 * 3 = 9).  Then the four sample movie titles are filled in.  Each movie title appears at least once, which should be sufficient for good recognition.

Compare this to the full cartesian product, which is 36 (3 * 3 * 4) utterances.  This is 4x larger than the default.  For this example, it's not a big deal, but the problem explodes when you have larger dictionaries and more slots.

```
DemoIntent	foo foo {star wars|MOVIE}
DemoIntent	bar foo {star wars|MOVIE}
DemoIntent	baz foo {star wars|MOVIE}
DemoIntent	foo bar {star wars|MOVIE}
DemoIntent	bar bar {star wars|MOVIE}
DemoIntent	baz bar {star wars|MOVIE}
DemoIntent	foo baz {star wars|MOVIE}
DemoIntent	bar baz {star wars|MOVIE}
DemoIntent	baz baz {star wars|MOVIE}
DemoIntent	foo foo {inception|MOVIE}
DemoIntent	bar foo {inception|MOVIE}
DemoIntent	baz foo {inception|MOVIE}
DemoIntent	foo bar {inception|MOVIE}
DemoIntent	bar bar {inception|MOVIE}
DemoIntent	baz bar {inception|MOVIE}
DemoIntent	foo baz {inception|MOVIE}
DemoIntent	bar baz {inception|MOVIE}
DemoIntent	baz baz {inception|MOVIE}
DemoIntent	foo foo {gattaca|MOVIE}
DemoIntent	bar foo {gattaca|MOVIE}
DemoIntent	baz foo {gattaca|MOVIE}
DemoIntent	foo bar {gattaca|MOVIE}
DemoIntent	bar bar {gattaca|MOVIE}
DemoIntent	baz bar {gattaca|MOVIE}
DemoIntent	foo baz {gattaca|MOVIE}
DemoIntent	bar baz {gattaca|MOVIE}
DemoIntent	baz baz {gattaca|MOVIE}
DemoIntent	foo foo {the matrix|MOVIE}
DemoIntent	bar foo {the matrix|MOVIE}
DemoIntent	baz foo {the matrix|MOVIE}
DemoIntent	foo bar {the matrix|MOVIE}
DemoIntent	bar bar {the matrix|MOVIE}
DemoIntent	baz bar {the matrix|MOVIE}
DemoIntent	foo baz {the matrix|MOVIE}
DemoIntent	bar baz {the matrix|MOVIE}
DemoIntent	baz baz {the matrix|MOVIE}
```

In my current app, I have dictionaries with 30+ values in them.  Combine that with 2 or 3 shortcuts in an utterance, 8 utterances per intent, and very quickly, I exceed the 250KB sample utterances limit at Amazon.  In fact, my utterances were more than 20MB!  Using the minimalistic technique resulted in about 40KB.  This would let me use many more values in my dictionaries, which might help me tune the recognition a little better.

BTW -- the code is designed that for each defined utterance in the Intent, it will ensure that there are enough generated utterances so that every value from every dictionary can be used at least once.  Technically, I think we only need to make sure that the dictionary values for a given slot are used once per *Intent*, not once per alexa-app defined utterance.  So I'm being a little wasteful here, but it kept my code changes to a minimum.

My code isn't quite as elegant as yours; there may be better ways to accomplish what I'm trying to do.  Feel free to reject the pull request if there's a better/cleaner way to do this.  But I believe this is a real need when you start dealing with real-world data where you need lots of sample inputs.

### DIFF
--- a/index.js
+++ b/index.js
@@ -119,6 +119,9 @@ alexa.app = function(name,endpoint) {
 	
 	// Persist session variables from every request into every response?
 	this.persistentSession = true;
+
+	// use the minimal set of utterances or the full cartesian product?
+	this.exhaustiveUtterances = false;
 	
 	// A catch-all error handler - do nothing by default
 	this.error = null;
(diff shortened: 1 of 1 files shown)


matt-kruse:
This is great, thank you. I have been wanting to add the same thing, but haven't had time. I will review your code changes when I get a chance and hopefully merge them in if it looks good to me. Thanks again!

mreinstein:
@jpriebe will you please submit this PR to https://github.com/mreinstein/alexa-utterances ?

We're splitting this out into its' own module and this seems like a valuable thing to include.

matt-kruse:
I just committed an update so alexa-app now uses alexa-utterances. Please do move this functionality to that module. Thanks. :)

PR #24, opened 2015-10-02
Upgrade Clear/Reprompt/Say methods to SSML

tvand7093:
Amazon just updated to their new SSML feature for speech output. Since I was playing with that new feature I thought I would update your API to use this. The `response.clear/reprompt/say` now get wrapped in the new `<speak></speak>` tags and the new speech type is applied.

I have been testing and playing with this all night (you can see my test project [here](https://github.com/tvand7093/Alexa-node)) and it seems to work! I don't really have time to try and update the README or anything, but hopefully this code change is useful for others.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,23 +15,23 @@ alexa.response = function() {
 	};
 	this.say = function(str) {
 		if (typeof this.response.response.outputSpeech=="undefined") {
-			this.response.response.outputSpeech = {"type":"PlainText","text":str};
+			this.response.response.outputSpeech = {"type":"SSMS","ssms": "<speak>" + str + "</speak>" };
 		}
 		else {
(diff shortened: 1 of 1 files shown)


colinsullivan:
+1

Is there ever a case where plain text output is necessary?

tvand7093:
I have no idea. From what I can see it is just optional. But i'm not positive on that. I couldn't find anything saying that it isn't just optional at this point. I suppose the only situation would be to safe on the string size being sent/received?

christianewillman:
+1, please please add SSML support.

tekkkon:
vote for supporting SSML

matt-kruse:
Resolved using code from @rickwargo in Commit #31

PR #31, opened 2015-11-20
adding support for SSML in responses

rickwargo:
I have added initial support for making SSML responses, converting each plain_text response to SSML. There is utility associated such that it knows to remove the SSML in a card, and can handle linking of speechOutputs such that there is only one outer <speak /> element.

This allows you to embed SSML in your response.say() calls and makes it easy to use that text in the response.card() call as it will strip out only the SSML.

I left .clear() alone as having an empty response in PLAIN_TEXT should not break anything. The first assigment will change it to SSML.

NOTE: One of the other things SSML.fromStr provides is a mapping of number literals to word phrases through the numbered module. Even though Alexa speaks the number literals, if it is at the end of the sentence, it will also say "dot" after reading the number. I believe the API states that the word should be used instead of the literal.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 var Promise = require('bluebird');
 var AlexaUtterances = require('alexa-utterances');
+var SSML = require('./to-ssml');
 
 var alexa={};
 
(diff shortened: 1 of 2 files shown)


matt-kruse:
Note - This change introduced a dependency on numbered, which wasn't in package.json as a dependency. Resolved in version 2.3.1

OverloadUT:
Although version 2.3.2 will take that module back out, haha!

PR #34, opened 2015-11-20
Trim spaces for utterances

rickwargo:
Using a construct like (|this) will have an utterance with a preceding space. This will remove that space. Note the only change is associated with commit https://github.com/matt-kruse/alexa-app/commit/7b80f3dc557ce5c787297eaafc4a36c59abcdb58

### DIFF
--- a/index.js
+++ b/index.js
@@ -319,7 +319,7 @@ alexa.app = function(name,endpoint) {
 		}
 		if (enableDebug) {
 			express.get(endpoint,function(req,res) {
-				res.render('test',{"json":self,"schema":self.schema(),"utterances":self.utterances(),self.customSlotTypes()});
+				res.render('test',{"json":self,"schema":self.schema(),"utterances":self.utterances(),"slot_types":self.customSlotTypes()});
 			});
 		}
 	};
--- a/index.js
+++ b/index.js
@@ -256,7 +256,26 @@ alexa.app = function(name,endpoint) {
 		};
 		return JSON.stringify(schema,null,3);
 	};
-	
+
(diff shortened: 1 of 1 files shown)


mutexkid:
Hi @rickwargo - i think we should close this branch. 
My thought with the custom slot types is that it's not going to be great including the content for the custom slots within the framework. Some of these lists can contain hundreds of items! Having them in text files to feed into the skills interface seems like the best option for now - until hopefully amazon provides us with a standard xml or json definition we can send for having the skill service share it's definition for the schema, slot types, utterances, and custom slots. 

I think the easiest thing might be to run with the suggestion made on the alexa-utterances repo from 
@simplydallas : just add a .replace(/\{\-\|/g, "{") to the output. I also think adding the trim() from this branch to fix the output would be an improvement to the output. 

Sound good? I'd like to create a new branch that brings the custom slot regex and the trim() statement forward, along with tests for checking it works.

GitTom:
@mutexkid @rickwargo 
> Some of these lists can contain hundreds of items! 

Thousands.  The current limit is 50K, and I saw someone complaining the forums about going over the limit.  So I guess I'm seconding mutexkid.

mutexkid:
Indeed - i used "hundeds" only as an illustration here - even schlepping 50 values into a javascript array only to have it spit back out again in the server seems to be a tad rediculous :D

rickwargo:
@mutexkid That sounds good to me. The real challenge is how to populate the Skill Interaction Model as you mentioned. Having those custom slot values in a consistent format is the first step; even if it is just by convention. I threw them in my model just to try to enforce that; I'm happy keeping them in text files.

PR #35, opened 2015-11-21
Added user to request in order to be able to get the accessToken

nsainaney:
Alexa Skills can now pair Amazon Echo accounts using OAuth. Added request.user in order to support this.


matt-kruse:
Resolved with Commit #37

PR #38, opened 2015-12-15
Add: Support for the LinkAccount card type

christianewillman:
I would rather add a `type` option to `#card`, but I don't want to add a fourth parameter. Users will have to be careful to not call `#card` after calling `#linkAccount`.

On the bright side, I guess this protects us if Amazon ever changes the means by which one invokes the account linking process.

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,6 +38,10 @@ alexa.response = function() {
 		this.response.response.card = {"type":"Simple","content":content,"title":title,"subtitle":subtitle};
 		return this;
 	};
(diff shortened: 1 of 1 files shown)


christianewillman:
```js
app.intent('MyAuthenticatedIntent', function(request, response) {
  const accessToken = request.data.session.user.accessToken;

  if (!accessToken) {
    response.linkAccount();
  } else {
    response.say('Welcome back, ' + lookupUser(accessToken).firstName);
  }
});

PR #43, opened 2016-01-05
Incorporate an SSML library for responses

MaxwellPayne:
Can now use an [SSML library](https://github.com/TheFuzzy/node-ssml) to allow more elegant buildup of SSML responses, as @rickwargo requested in a TODO comment in #31. 

Building up an SSML response looks like:

``` javascript
var message = new ssml().say('The time is')
    .say({
        'text': '01:59:59',
        'interpretAs': 'time',
        'format': 'hms24'
    }).break(500);

app.say(message)

// Response is:
'<speak>The time is<say-as interpret-as="time" format="hms24">01:59:59</say-as><break time="500ms"/></speak>'

// You can still use strings in app.say(), and you can mix them with SSML
app.say('Is my shift over yet?')

// Response is now:
'<speak>The time is<say-as interpret-as="time" format="hms24">01:59:59</say-as><break time="500ms"/> Is my shift over yet?</speak>'
```

I also started a test suite using [tap](https://github.com/tapjs/node-tap). Right now it only tests changes introduced on this PR, but it could be a good starting point for tests in the future as this repo grows in popularity.


MaxwellPayne:
**Note:** right now the `package.json` has a dependency on [my fork](https://github.com/MaxwellPayne/node-ssml) instead of the SSML library's original repo. Though the original repo is hosted on npm, it has a few bugs that render the `<say-as>` tag useless. And because this is one of the most important tags in my opinion, I decided depending on my own fork would be best until @TheFuzzy fixes these issues and resubmits to npm.

matt-kruse:
This doesn't need to be integrated into alexa-app. You can include the ssml library in your own app and build the response, then feed that to alexa-app, right?

BechtelCanDoIt:
But you need ssml to include audio clips and to have Alexa spell things
out. Why would you want that in a framework?

OverloadUT:
I'm inclined to agree with Matt - this app should simply accept SSML and pass it along to the output, and it's up to the developer what library they want to use. I don't see any benefit to prescribing which library the developer must use

MaxwellPayne:
I'm not sure how many SSML libraries there are out there. If the one I used is "the one," I think it makes sense to support it natively. But if there are a number of different options then we should do as Matt said and keep it minimal.

dblock:
I'm going to close this given the discussion above, LMK if you feel otherwise.

PR #44, opened 2016-01-06
Adds support for an optional "_default" intent, which will be used when the supplied intent is not defined

OverloadUT:
Pretty simple. Allows setting up a "fallback" intent by registering it with the name "_default".

### DIFF
--- a/index.js
+++ b/index.js
@@ -203,6 +203,11 @@ alexa.app = function(name,endpoint) {
 				if (!response.resolved) {
 					if ("IntentRequest"===requestType) {
 						var intent = request_json.request.intent.name;
+						if (typeof self.intents[intent]==="undefined" && typeof self.intents["_default"]!=="undefined") {
+							// If this intent doesn't have a handler, but there is a _default handler defined,
+							// send it to the _default handler
+							intent = "_default";
+						}
 						if (typeof self.intents[intent]!="undefined" && typeof self.intents[intent]['function']=="function") {
 							if (false!==self.intents[intent]['function'](request,response)) {
 								response.send();


matt-kruse:
Why would you need a default intent? You know every possible intent already, right?

OverloadUT:
Admittedly, it's only for a few corner cases:
1. It's possible that you could have multiple people configure Skills that point to your single endpoint, and in those cases they might send Intents that you don't have configured and you want a friendly message (or functionality).
2. In my case, I use a state machine to do a whole bunch of stuff before I set up the app with intent() calls, and in some of my states I just want all of the invalid intents to have the same handler. I could do this differently of course but it'd be more convenient this way.

But from a philosophical perspective it just feels right to me to have a way to handle unknown intents in the same way that you can handle the known ones.

MaxwellPayne:
Is there a such thing as an "unknown" intent? If you invoke Alexa with some sort of nonsense intent not registered in the interaction model, what happens?

mreinstein:
> Why would you need a default intent?

Valuable for seeing what kinds of things users throw at your skill, and incorporating that feedback into future changes.

dblock:
Alexa doesn't have support for a "default" intent, so there's no predictable way to invoke it. I understand why you would want a fallback where the scheme doesn't define a certain intent and it gets called, but it seems like not a realistic scenario and feels more like a side-effect. I'm going to close this, but I could be getting it wrong, so please feel free to add more info and reopen.

nebrekker:
I'm glossing over this discussion but I have noticed that, while it's possible for slots to have any values (or no value at all), Alexa seems determined to slot your utterance into a defined intent. There seems to be no such thing as a default or fall through intent. For my adventure game I have actually created a dummy intent seeded with nonsensical utterances.  Admittedly it's a hack, but it gives the intent schema something to fall back on if someone says something entirely outside the scope of my game. To the extent that it can return an error message to the user (as opposed to "randomly" performing the closest matching intent utterance--which were often spoilers or deadly accidents to the player), this is the most natural course of action I have found and works in alexa-app as designed.

Now about those slots. There may be another thread and if I find one I'll comment. I get 

missing intent in request: actionItem [TypeError: Cannot read property 'value' of undefined]

All the time and was about to raise an issue, but they seem to have no consequence other than noise on the console, and might be useful.  I would just note that missing slot values (at present) seems to be a natural phenomenon and something I actually use strategically to guide my intents. If I would ask for anything it would be a less "error" sounding informational error, but that would be nothing more than "nice to have."

PR #46, opened 2016-01-08
added missing numbered module dependency

mreinstein:
@matt-kruse alexa-app@2.3.0 uses the `numbered` module, but it's not specified in the package.json file. This fixes it. Should publish a new version to npm asap because the current version is totes busted. :-/

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   "license": "MIT",
   "dependencies": {
     "alexa-utterances": "^0.1.0",
-    "bluebird": "^2.10.2"
+    "bluebird": "^2.10.2",
+    "numbered": "^1.0.0"
   }
 }


matt-kruse:
Wow, I'm not sure how that happened. Thanks for the quick fix.

mreinstein:
At some point we should add some basic tests based on https://www.npmjs.com/package/tap and add this as a prepublish step in package.json. Won't totally fix the problem but it'd give some nice fast boundary checks.

matt-kruse:
Agreed, would love to have tests, especially since this framework is getting increasingly popular.
BTW, this missing dependency was introduced in the SSML merge, which is why I missed it. I need to pay closer attention when merging other peoples' pull requests...

mreinstein:
I'm working on another PR with some fledgling tests and the prepublish hook. I'll have something for ya in a few mins.

mreinstein:
see https://github.com/matt-kruse/alexa-app/pull/48

I went with tap because it's very lightweight and minimal. I tend to prefer tiny testing libraries rather than large frameworks.

PR #48, opened 2016-01-08
integrated tap tests, setup npm prepublish hook

mreinstein:


mutexkid:
Hi @mreinstein, thanks for the pull request. alexa-app is running with mocha and chai as the test stack, which means tap tests won't be supported. I agree adding comprehensive test coverage is a great idea and any contributions on improving coverage using mocha and chai would be a great addition. We will also be getting a travis integration set up soon. Marking this as closed.

PR #54, opened 2016-03-30
adding standard card type

jejernig:
supporting standard card type with images


HynesIP:
Oop. Also, the key for standard card type is text: instead of content: .

PR #55, opened 2016-03-30
Adding standard card type 

jejernig:
Allowing for the use of images in cards

### DIFF
--- a/index.js
+++ b/index.js
@@ -42,6 +42,9 @@ alexa.response = function() {
 		this.response.response.card = {"type":"Simple","title":title,"content":SSML.cleanse(content),"subtitle":subtitle};
 		return this;
 	};
+	this.standardCard = function(title, content, smallImageUrl, largeImageUrl){
+		this.response.response.card = {"type": "Standard", "title": title, "content":SSML.cleanse(content), "image": {"smallImageUrl" smallImageUrl, "largeImageUrl": largeImageUrl} };
+	}
 	this.linkAccount = function() {
 		this.response.response.card = {"type":"LinkAccount"};
 		return this;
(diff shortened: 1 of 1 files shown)


mutexkid:
1. not returning this will break the fluent interface we see with the `standardCard` method
2. I'm questioning whether having many different methods for displaying different types of cards is a good solution here. I think a better approach here might be to accept a new Card object instead? 
That way, as new types of cards are added, the api doesn't continue to expand, and instead we can offer new card types.

mutexkid:
@radishmouse any thoughts on this?

doapp-ryanp:
I like the card obj vs lots of methods.  The things you can do with cards will have to expand in the future, and your proposed design will be more resilient to this.

mutexkid:
Ok, thanks for weighing in on the design we should go with here.  Closing in favor of keeping `card` method with an eye to adding an OO approach to supporting the different card objects Skills Kit API currently supports. https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#card-object

The test coverage milestone is currently a priority, and once completed the new cards support can be addressed. I've added a milestone for this work to make sure we don't lose sight of the need for this work!

doapp-ryanp:
I should be able to write the code for this.  I could prob pull it off yet this afternoon / next few days depending on how quick we nail down the high level arch.

@mutexkid 

- Do you want me to do it (or is this core enough that you'd rather do it)?
- If so, whats the best place to present my proposed design before I code it up? https://github.com/matt-kruse/alexa-app/issues/53 ? Or do you want me to make a new issue that doesn't just tackle images, but broader support for all types (`Simple`,`Standard`,`LinkAccount`)

doapp-ryanp:
I decided to go ahead and implement it. PR above. Should be pretty future proof without needing to write much code.  Respond to https://github.com/matt-kruse/alexa-app/pull/59 if you want me to change something

PR #59, opened 2016-05-04
full card type support

doapp-ryanp:
With backwards compatibility support

Implements https://github.com/matt-kruse/alexa-app/issues/53 and https://github.com/matt-kruse/alexa-app/pull/55

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,6 @@ Network Trash Folder
 Temporary Items
 .apdisk
 /coverage
+
+#IDE
+.idea
--- a/README.md
+++ b/README.md
@@ -106,7 +106,9 @@ response.clear()
 response.reprompt(String phrase)
 
 // Return a card to the user's Echo app
-response.card(String title, String content)
+// For Object definition @see https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#card-object
+// Still supports card(String title, String content) for backwards compat of type "Simple"
+response.card(Object card)
 
 // Return a card instructing the user how to link their account to the skill.
 // This internally sets the card response.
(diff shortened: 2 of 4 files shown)


doapp-ryanp:
I tested every combination except `LinkAccount`:

-  Simple, Standard (both with no title)
-  Standard with images

I have also added some basic validation logic to help developers out as improper object shape errors are hard to track down.

doapp-ryanp:
A bit un-related, why are you not using ES6 features in the node LTS release (4.3.x)?

mutexkid:
Hi @doapp-ryanp -  i like where this is headed!
1. can we turn off the "autoaligning" with the function names? Just revert those back to how they were. 
2. we have a test suite for the project now. Rebase onto master and run the tests using `npm test` to make sure everything still works. (test suite needs more work!)
3 adding some tests for the new card behavior - I've got a file on master called https://github.com/matt-kruse/alexa-app/blob/master/test/test_alexa_app_cards.js that might be a good place to start!

Do you think we should aim to throw an exception if the fields the card needs to work correctly based on its type are missing?

mutexkid:
re: ES6 - there's a milestone for porting to ES6. Once we have good test coverage we can get that underway.

doapp-ryanp:
1. Done. Removed my formatting
1. Done. All tests pass
1. I'll work on adding some test cases based on your response to below

Re: throwing Exceptions.  I didn't do this because you don't current throw any Errors that are not caught within alexa-app/index.js.  Said another way, your module does not currently bubble up exceptions, and I didn't want to break that architecture.

The other reason I didn't bubble up, is I kinda made the assertion that the skill should not "break" if the card is invalid.  The intent response will still have speech, not sure if people would want no audio if card has an issue.  My thought was a developer would find out the 1st unit test if they called `.card()` incorrectly.

With that said, I'm cool bubbling up Exceptions, but I'd prefer to defer to you / your main team to implement this given that you will need to come up with a clearly conveyed design for when an Error is thrown and what type of Error is thrown (I'd propose you come up with your own Exception class that extends Error).

What you think?

mutexkid:
Yeah, i think its definitely a valid criticism to point out the way we are handling errors should be consistent. I do think we should reconsider how that's being done currently, but we can defer to when we're in a better place to do so (test coverage before any refactoring!). 
Let's hold off on enforcing some constraints on the object that's expected right now until we can look at that more carefully. 👍 Add some tests for the card object returning the correct response and we'll move forward.

doapp-ryanp:
Yea sorry didn't to come off harsh there re: no exception bubbling. was just more of an observation.

OK I'll add some tests now

mutexkid:
re: harsh, not at all. I think carefully considering the error handling needs to be done, and critiquing how it is put together is needed. perhaps as we move to es6 that might be a natural time to rethink how we are handling them?

doapp-ryanp:
Yea rolling it into your ES6 makes sense to me.  Bluebird promise lib has some really nice metaphors for handling throwing and catching of errors (and custom error classes).

doapp-ryanp:
Sorry I made the assertion that you'd be going with a promise arch based on bluebird in your ES6 re-write. That wold be my recommendation.

mutexkid:
re: promises - yeah, we will be sticking with bluebird as our promise implementation. 
Cleaning up how we're working with promise behavior as we have it defined right i'd like to see happen. 
Right now, the returning of true or false within intent handlers to indicate we want promises or not i believe we can improve upon. We'll get there!

doapp-ryanp:
OK testcases added

mutexkid:
🎊 Great - thanks for adding tests. Merged!

doapp-ryanp:
Great thanks. Can you npm publish a new version? I'd like to use it :)

mutexkid:
I'm unable to currently. @matt-kruse can you publish a new release of alexa-app to npm?

matt-kruse:
Can you bump the version? I'll update npm from the latest master when that's done.

matt-kruse:
Published 2.3.3 to npm

PR #66, opened 2016-05-14
Tests to cover clearSession and reprompt in shouldEndSession

rickwargo:
Added three tests to cover response.clearSession() and shouldEndSession(bool, reprompt).


mutexkid:
💯

PR #69, opened 2016-05-23
bump version

mutexkid:
@matt-kruse let's release a new version to npmjs.org of the library, including the recent added support for the custom slot types via `alexa-utterances`.


matt-kruse:
Published 2.3.4 to npm.

mutexkid:
🎊

PR #70, opened 2016-06-03
Add state to callback functions

mattotodd:
I am using this package to write [node-red bindings for alexa](https://github.com/mattotodd/node-red-contrib-alexa).  

The issue is that the way that [node-red](http://nodered.org/) processes messages,  it can not respond to a http request until the end of a chain of nodes, and then it uses the express response object to respond.

These nodes do not know about one another, they only pass a message (javascript object) down the chain, and the express response object needs to be passed down that chain.

HTTP In --->  Do Some Processing --> Add Alexa Say Command --> HTTP Out

The way the library was currently written, I couldn't find a way to pass along the express response object.
So for now I'm including it it my package with the changes, but I'd rather include `alexa-app` as a dependency 

The usage within my project can be found [here](https://github.com/mattotodd/node-red-contrib-alexa/blob/master/alexa.js#L81)

### DIFF
--- a/index.js
+++ b/index.js
@@ -222,7 +222,7 @@ alexa.app = function(name, endpoint) {
   this.sessionEnded = function(func) {
     self.sessionEndedFunc = func;
   };
-  this.request = function(request_json) {
+  this.request = function(request_json, state) {
     return new Promise(function(resolve, reject) {
       var request = new alexa.request(request_json);
       var response = new alexa.response();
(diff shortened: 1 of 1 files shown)


mattotodd:
a love here?

dblock:
Sorry for the delay, we're back on track here. I think this is valuable. It would need tests and CHANGELOG to get this in. Care to do it @mattotodd?

kevingilbert100:
+1

PR #79, opened 2016-08-25
Update README.md

badcab:
Only clear the response if an exception was thrown


dblock:
I want to merge this, can we get the spacing right, like `if[space](...)` and such?

dblock:
Merged via https://github.com/matt-kruse/alexa-app/commit/82321272fdf9c23074adb921675c42ebf765c781.

PR #83, opened 2016-09-18
keep multiple newlines in card output

kielni:
Cards can use multiple newlines for formatting, but https://github.com/matt-kruse/alexa-app/blob/master/index.js#L91 reduces multiple newlines to one.  

This change skips the newline compression when cleansing SSML for card output.

Here's a sample card formatted with newlines:

multiline_card

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,7 +88,7 @@ alexa.response = function() {
 
     // remove all SSML to keep the card clean
     clenseAttrs.forEach(function(idx) {
-      oCard[idx] = SSML.cleanse(oCard[idx]);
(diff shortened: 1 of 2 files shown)


dblock:
This makes sense. I would merge this with tests and a CHANGELOG entry, please.

dblock:
Bump @kielni want to finish this?

kevingilbert100:
This is a huge priority, can we update the changelog? I'll go ahead and throw in a different PR today with the same change and the changelog update.

dblock:
Looking forward to your PR @kevingilbert100.

PR #88, opened 2016-11-14
AudioPlayer functionality

wschaeferiii:
Added Audioplayer functionality to the alexa-app library.

Simply write `response.audioPlayerPlay(url, token, playBehavior, offsetInMilliseconds, expectedPreviousToken)` and you can play long form audio URL requests with a given url and token.  `playBehavior` defaults to `"ENQUEUE_ALL"`, and `offsetInMilliseconds` defaults to 0, and `expectedPreviousToken` defaults to `undefined`.

Also, added is `response.audioPlayerStop()` and `response.audioPlayerClearQueue(clearBehavior)` where `clearBehavior` defaults to `"CLEAR_ALL"` .

The request's `context` object is required in long form audio requests since the session object is not included once the player begins playing, but is copied from the request's `session` object.

Issue #78, opened 2016-08-24
Error when uploading alexa-app to aws lamba


siddhartha-lahiri:
Hi,

I am trying to create a basic alexa-app and upload the zip file to lambda.
I have used generator-lambda to create a basic skeleton.

This is the error I keep on getting

```
{
  "errorMessage": "Cannot read property 'new' of undefined",
  "errorType": "TypeError",
  "stackTrace": [
    "new alexa.request (/var/task/node_modules/alexa-app/index.js:149:29)",
    "/var/task/node_modules/alexa-app/index.js:227:21",
    "tryCatcher (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/util.js:26:23)",
    "Promise._resolveFromResolver (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/promise.js:480:31)",
    "new Promise (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/promise.js:70:37)",
    "request (/var/task/node_modules/alexa-app/index.js:226:12)",
    "handler (/var/task/node_modules/alexa-app/index.js:357:10)"
  ]
}
```

PaeP3nguin:
Try your skill out anyways. I find that the error only occurs on upload and testing in Lambda, but the skill itself works just fine. :/

OverloadUT:
That is likely because you haven't changed the Test payload in Lambda. The default one is not an Alexa payload. You can change it to a simple LaunchRequest or something as a good way to sanity check that the uploaded codeworks.

dblock:
Closing this, please let us know if you're still having issues @siddhartha-lahiri.

fremail:
@dblock I got this error.

I have an app uploaded to Lambda.

The error is caused on getting `AudioPlayer` events. I got it twise: one for `AudioPlayer.PlaybackStarted` and one for `AudioPlayer.PlaybackFinished` I suppose.

```
{
    "errorMessage": "Cannot read property 'new' of undefined",
    "errorType": "TypeError",
    "stackTrace": [
        "new alexa.request (/var/task/node_modules/alexa-app/index.js:219:27)",
        "/var/task/node_modules/alexa-app/index.js:317:18",
        "tryCatcher (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/util.js:26:23)",
        "Promise._resolveFromResolver (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/promise.js:483:31)",
[...]

PR #91, opened 2016-12-19
Fix "Cannot read property 'new' of undefined" init error

fremail:
The error is caused on getting `AudioPlayer` events. I got for these events: `AudioPlayer.PlaybackStarted`, `AudioPlayer.PlaybackNearlyFinished` and `AudioPlayer.PlaybackFinished`.

```
{
    "errorMessage": "Cannot read property 'new' of undefined",
    "errorType": "TypeError",
    "stackTrace": [
        "new alexa.request (/var/task/node_modules/alexa-app/index.js:219:27)",
        "/var/task/node_modules/alexa-app/index.js:317:18",
        "tryCatcher (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/util.js:26:23)",
        "Promise._resolveFromResolver (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/promise.js:483:31)",
        "new Promise (/var/task/node_modules/alexa-app/node_modules/bluebird/js/main/promise.js:71:37)",
        "request (/var/task/node_modules/alexa-app/index.js:316:10)",
        "handler (/var/task/node_modules/alexa-app/index.js:448:8)"
    ]
}
```

As Amazon [wrote](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#request-body-parameters):
> The session object is included for all standard requests, but it is not included for AudioPlayer or PlaybackController requests.

Additional info regarding to this error you can find in PR #88 and Issue #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ app.intent("number", {
     "slots": { "number": "NUMBER" },
     "utterances": ["say the number {1-100|number}"]
   },
-  function(request, response) {
+  function(request, response, session) {
     var number = request.slot("number");
     response.say("You asked for the number " + number);
   }
@@ -57,15 +57,6 @@ String request.type()
 
 // return the value passed in for a given slot name
 String request.slot("slotName")
-
-// check if you can use session (read or write)
-Boolean request.hasSession()
-
-// return the value of a session variable
-String request.session("attributeName")
-
-// session details, as passed by Amazon in the request
-request.sessionDetails = { ... }
 
 // the raw request JSON object
 request.data
@@ -99,14 +90,6 @@ response.linkAccount()
 // you can optionally pass a reprompt message
 response.shouldEndSession(boolean end [, String reprompt] )
 
-// check if you can use session (read or write)
-Boolean response.hasSession()
-
-// set a session variable
-// by defailt, Alexa only persists session variables to the next request
-// the alexa-app module makes session variables persist across multiple requests
-response.session(String attributeName, String attributeValue)
-
 // send the response to the Alexa device (success)
 // this is not required for synchronous handlers
 // you must call this from asynchronous handlers
@@ -121,6 +104,23 @@ response.fail(String message)
 response.say("OK").send()
 ```
 
+## session
+```javascript
+// check if you can use session (read or write)
+Boolean session.isAvailable()
+
+// set a session variable
+// by defailt, Alexa only persists session variables to the next request
+// the alexa-app module makes session variables persist across multiple requests
+session.set(String attributeName, String attributeValue)
+
+// return the value of a session variable
+String session.get(String attributeName)
+
+// session details, as passed by Amazon in the request
+session.details = { ... }
+```
+
 # Request Handlers
 
 Your app can define a single handler for the `Launch` event and the `SessionEnded` event, and multiple intent handlers.
@@ -128,7 +128,7 @@ Your app can define a single handler for the `Launch` event and the `SessionEnde
 ## LaunchRequest
 
 ```javascript
-app.launch(function(request, response) {
+app.launch(function(request, response, session) {
   response.say("Hello World");
   response.card("Hello World", "This is an example card");
 });
@@ -139,11 +139,11 @@ app.launch(function(request, response) {
 Define the handler for multiple intents using multiple calls to `intent()`. Intent schema and sample utterances can also be passed to `intent()`, which is detailed below. Intent handlers that don't return an immediate response (because they do some asynchronous operation) must return false. See example further below.
 
 ```javascript
-app.intent("buy", function(request, response) {
+app.intent("buy", function(request, response, session) {
   response.say("You bought a " + request.slot("item"));
 });
 
-app.intent("sell", function(request, response) {
+app.intent("sell", function(request, response, session) {
   response.say("You sold your items!");
 });
 ```
@@ -151,7 +151,7 @@ app.intent("sell", function(request, response) {
 ## SessionEndRequest
 
 ```javascript
-app.sessionEnded(function(request, response) {
+app.sessionEnded(function(request, response, session) {
   // cleanup the user's server-side session
   logout(request.userId);
   // no response required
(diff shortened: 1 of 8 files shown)


dblock:
This needs a test and a CHANGELOG entry, please?

index.js
```diff
@@ -145,6 +145,14 @@ alexa.request = function(json) {
       return null;
     }
   };
+  // session is not included for AudioPlayer or PlaybackController requests
```
dblock: Maybe add the https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#request-body-parameters link here too?

dblock:
Before we merge this - since the Alexa requests don't include session objects for these audio ones, maybe we shouldn't be making a session object to work around that, but being more defensive? Otherwise we're creating something not supported by the platform, no?

fremail:
Actually I agree with you, we shouldn't provide session interface in case of its absence.

We can set `request.userId`, `request.applicationId`, etc. from request context object. But what will we do with `request.session()` function? Will we define it in that case?

dblock:
Maybe we should throw an error when you access `session()` in this case?

index.js
```diff
@@ -109,18 +114,26 @@ alexa.response = function() {
...
-      return this.response.sessionAttributes[key];
+    if (this.hasSession()) {
+      if (typeof val == "undefined") {
+        return this.response.sessionAttributes[key];
+      } else {
+        this.response.sessionAttributes[key] = val;
+      }
     } else {
-      this.response.sessionAttributes[key] = val;
+      throw "NO_SESSION";
```
dblock: Is this a pattern we use elsewhere? Maybe throw a more helpful message?

dblock: Oh scratch that, I see it below.

index.js
```diff
@@ -180,6 +204,9 @@ alexa.app = function(name, endpoint) {
     "NO_LAUNCH_FUNCTION": "Try telling the application what to do instead of opening it",
     // When a request type was not recognized
     "INVALID_REQUEST_TYPE": "Error: not a valid request",
+    // When a request and response don't contain session object
+    // https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#request-body-parameters
+    "NO_SESSION": "There is not session to store or get values",
```
dblock: Maybe "this request doesn't support session attributes" or something like that?

fremail:
Sorry I didn't write tests, dunno how to do it.
And it seems existed tests failed because the test requests don't contain `context` object, though should have it. [Link](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#context-object)

The idea of my changes:
Deny using `session` in `request` and `response` objects, so all functions using it will throw "NO_SESSION". To check if you can use those functions you can call `request.hasSession()` or `response.hasSession()`

dblock:
See my comment above wrt `hasSession` and I can help you with tests. Start by having the tests that exist pass (`npm test`), then try to add a test that checks that given one of those requests  that causes a session not to exist you can get a response. Then add a test that lands in the `throw` part with `NO_SESSION`.

dblock:
And thanks for the hard work! Hang in there we'll get this to a mergeable state soon.

index.js
```diff
@@ -3,15 +3,20 @@ var AlexaUtterances = require("alexa-utterances");
 var SSML = require("./to-ssml");
 var alexa = {};
 
-alexa.response = function() {
+alexa.response = function(hasSession) {
```
fremail: As I see your idea:
```
request.session = {
    this.get = function(key) {...};
    this.set = function(key, value) {...};
    this.clear = function() {...};
    this.details = {...};
    this.attributes = {...};
    this.isNew = bool;
}
```

Just don't understand why we need to store the same data in `this.sessionDetails` and in properties of `this`.

BTW, these changes will break compatibility.

And we need to edit `response.response` from the `session` object. Don't sure how we can do that.

dblock: I haven't actually written the code, give it a shot. Obviously backwards compat breaking is bad, maybe we can minimize that once we see what it looks like.

fremail: Maybe we need to move `session` outside of `request` and `response` since it's an independent object not related to request or response instantly. 
I'd even say it's between them.
So we can provide it as an alone object: 
```
app.pre = function(request, response, type, session) {
  ...
};

app.intent("sampleIntent", {
    "slots": {...},
    "utterances": [...]
  },
  function(request, response, session) {
    ...
  }
);
```
And call `session` methods from existed methods of `request` and `response` for backwards compatibility.

fremail:
So I separate `session` object from `alexa.request` and `alexa.response` - now it's `alexa.session`.

These methods work for 100% backwards compatibility, but marked as deprecated:
* `request.session()`
* `response.session()`
* `response.clearSession()`
The same for these properties:
* `request.sessionDetails`
* `request.sessionId`
* `request.sessionAttributes`
* `request.isSessionNew`

Now it needs to do `response.prepare()` to save `sessionAttributes` into `response.response` object.

README.md
```diff
@@ -115,14 +104,31 @@ response.fail(String message)
 response.say("OK").send()
 ```
 
+## session
+```javascript
+// check if you can use session (read or write)
+Boolean session.isAvailable()
```
dblock: It's a little strange to ask the session whether it's available. Why wouldn't this be on `request.hasSession()`? And then I think we don't need to change the first `function` because `request.session` would be a thing?

fremail: It seems we are tolking about different. `request.session()` and `response.session()` are in use. If we change them (to object, for example), then current apps using alexa-app might be broken.

fremail: Do you want to make a gettter in `request` object for getting `session` object so we don't need to pass `session` as a last argument to handlers?

dblock:
The code looks perfect. Just need two tests: for the loop that involves no session and another that ends up in that NO_SESSION exception being thrown.

dblock:
Lets also add an UPGRADING.md document that describes the methods that have been deprecated. For a good example see https://github.com/ruby-grape/grape/blob/master/UPGRADING.md

dblock:
Merged squashed. You'll have to fix/rewind/rebase your master branch (you should start working on branches). See http://code.dblock.org/2015/08/31/getting-out-of-your-first-git-mess.html that might be helpful.

fremail:
Woohoo! 
Will read it, thanks

fremail:
I cleaned my master branch, now it has only one commit for this PR, but the commit is after merge of this PR :(

dblock:
@fremail 

```
git checkout master
git checkout -b backup-master
git checkout master
git reset HEAD~50 --hard
git pull upstream master
git push origin master -f
```

PR #92, opened 2016-12-23
AudioPlayer functionality (Continue PR #88)

fremail:
Continuing PR #88 I added `app.audioPlayer(event, func)`
It allows to add an event handler to these AudioPlayer events:
* PlaybackStarted
* PlaybackFinished
* PlaybackStopped
* PlaybackNearlyFinished
* PlaybackFailed

Also this function supports async responses (similar to `intent()`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,33 @@ app.sessionEnded(function(request, response) {
 });
 ```
 
+## AudioPlayer Event Request
+
+Define the handler for multiple events using multiple calls to `audioPlayer()`. You can define only one handler per event. Event handlers that don't return an immediate response (because they do some asynchronous operation) must return false. 
+You can define handlers for these events:
+* PlaybackStarted
+* PlaybackFinished
+* PlaybackStopped
+* PlaybackNearlyFinished
+* PlaybackFailed
+
+See example further below.
+
+```javascript
+app.audioPlayer("PlaybackNearlyFinished", function(request, response) {
+  // immediate response
+  response.audioPlayerPlay("https://next-song-url", "some-token", "ENQUEUE");
+});
+app.audioPlayer("PlaybackFinished", function(request, response) {
+  // async response
+  getNextSongFromDB(function(url, token) {
+    response.audioPlayerPlay(url, token, "ENQUEUE");
+    response.send();
+  });
+  return false;
+});
+```
+
 # Execute Code On Every Request
 
 In addition to specific event handlers, you can define functions that will run on every request.
--- a/README.md
+++ b/README.md
@@ -92,6 +92,17 @@ response.card(Object card)
 // this internally sets the card response
 response.linkAccount()
 
+// play audio stream (send AudioPlayer.Play directive) @see https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/custom-audioplayer-interface-reference#play-directive
+// skill supports stream(String url, String token, String expectedPreviousToken, Integer offsetInMilliseconds)
+response.audioPlayerPlayStream(String playBehavior, Object stream)
+
+// stop playing audio strem (send AudioPlayer.Stop directive)
+response.audioPlayerStop()
+
+// clear audio player queue (send AudioPlayer.ClearQueue directive)
+// clearBehavior is "CLEAR_ALL" by default
+response.audioPlayerClearQueue([ String clearBehavior ])
+
 // tell Alexa whether the user's session is over; sessions end by default
 // you can optionally pass a reprompt message
 response.shouldEndSession(boolean end [, String reprompt] )
@@ -186,7 +197,13 @@ The following example will return `play` directive with a next audio on `AudioPl
 ```javascript
 app.audioPlayer("PlaybackNearlyFinished", function(request, response) {
   // immediate response
-  response.audioPlayerPlay("https://next-song-url", "some-token", "ENQUEUE", "previous token");
+  var stream = {
+    "url": "https://next-song-url",
+    "token": "some_token",
+    "expectedPreviousToken": "some_previous_token",
+    "offsetInMilliseconds": 0
+  };
+  response.audioPlayerPlayStream("ENQUEUE", stream);
 });
 ```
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -167,6 +167,33 @@ app.sessionEnded(function(request, response) {
 });
 ```
 
+## AudioPlayer Event Request
+
+Define the handler for multiple events using multiple calls to `audioPlayer()`. You can define only one handler per event. Event handlers that don't return an immediate response (because they do some asynchronous operation) must return false.
```
dblock: Maybe add a link to audio player support on some amazon site to explain what this is somewhere?

README.md
```diff
@@ -167,6 +167,33 @@ app.sessionEnded(function(request, response) {
 });
 ```
 
+## AudioPlayer Event Request
+
+Define the handler for multiple events using multiple calls to `audioPlayer()`. You can define only one handler per event. Event handlers that don't return an immediate response (because they do some asynchronous operation) must return false. 
+You can define handlers for these events:
```
dblock: for the _following_ events ...

And space things out a bit, add an empty line above and below.

README.md
```diff
@@ -167,6 +167,33 @@ app.sessionEnded(function(request, response) {
...
+* PlaybackStarted
+* PlaybackFinished
+* PlaybackStopped
+* PlaybackNearlyFinished
+* PlaybackFailed
+
+See example further below.
```
dblock: I would say "The following example [explain what this does]."

README.md
```diff
@@ -167,6 +167,33 @@ app.sessionEnded(function(request, response) {
...
+
+```javascript
+app.audioPlayer("PlaybackNearlyFinished", function(request, response) {
+  // immediate response
+  response.audioPlayerPlay("https://next-song-url", "some-token", "ENQUEUE");
+});
+app.audioPlayer("PlaybackFinished", function(request, response) {
+  // async response
+  getNextSongFromDB(function(url, token) {
```
dblock: Maybe add something that actually works as long as it's a 1-liner, like fetch some audio clip from some public website?

index.js
```diff
@@ -131,7 +133,67 @@ alexa.response = function(session) {
     this.sessionObject.clear(key);
     return this;
   };
-
+  this.audioPlayerPlay = function (url, token, playBehavior, offsetInMilliseconds, expectedPreviousToken) {
```
dblock: Which parts are required? Sounds like it's url and playBehavior. So maybe it shouldn't be method that takes 5 parameters but only those 2 and additional options that just get merged? This way you don't need the `switch(arguments.length)` at all.

index.js
```diff
@@ -131,7 +133,67 @@ alexa.response = function(session) {
...
+  this.audioPlayerStop = function () {
+    var audioPlayerDirective = {
+      "type": "AudioPlayer.Stop"
+    };
+    self.response.response.directives.push(audioPlayerDirective);
+    return this;
+  };
+  this.audioPlayerClearQueue = function (clearBehavior) {
+    var audioPlayerDirective;
+    if (arguments.length === 0) {
```
dblock: There should be a cleaner way to do this one without the big if. 

```js
audioPlayerDirective = {}
if () { ... }
self.response....
```

index.js
```diff
@@ -313,9 +412,10 @@ alexa.app = function(name, endpoint) {
         }
       };
       try {
+        var context = request.context();
         var requestType = request.type();
         if (typeof self.pre == "function") {
-          self.pre(request, response, requestType);
+          self.pre(request, response, requestType, context);
```
dblock: I think `pre` shouldn't take the 4th parameter since `request.context` is a thing. You can access that inside `pre`.

fremail: It was added by @wschaeferiii. Really I don't know reasons to add it.

index.js
fremail: Agree. I'll rewrite this function

index.js
dblock: Lets remove code that we don't know why it's there :)

index.js
fremail: `expectedPreviousToken` is required only in case of `playBehavior` is ENQUEUE. Otherwise we don't need it. 
Other params are required in all responses: [look at the table there](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/custom-audioplayer-interface-reference#play-directive)

We can try to assign default values to the arguments and make as you wrote, but I don't sure it'd be convenient.

README.md
fremail: I don't have ideas what it could be

README.md
fremail: Added this link: https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/custom-audioplayer-interface-reference#audioplayer-requests

index.js
dblock: I still think this is not future-proof. Every time Amazon adds something to audioplay we'll be throwing more parameters in here. What do you think about leaving this to the user and not being too smart about it?

```js
this.audioPlayerPlay = function (playBehavior, audioItem) {

}

this.audioPlayerPlayStream = function (playBehavior, stream) {
  
}
```

fremail:
Interesting.. when you expand readme file, test coverage decreases.

dblock:
Merge, excellent work @fremail and @wschaeferiii.

README.md
wschaeferiii: Hey @fremail , thank you for fixing my sloppy code and adding the AudioPlayer Event Request feature! 

I would love to use the feature, however having issues issues with my app listening to AudioPlayer events.  When the Alexa sends a audioplayer request (I can log the request_json, so I know its sent), the try / catch block with the `var requestType = request.type()` isn't even firing in my code.

Am I missing something here?

dblock: @wschaeferiii I know it's a continuation of this code, but just open a new issue for any bug you encounter against 2.4.0, hopefully with a test that reproduces the problem.

fremail: @wschaeferiii it's hard to suspect why it doesn't work for you. Please create an issue with a code example. Or point me to the issue please.

dblock:
Btw, just used this in https://github.com/artsy/elderfield/pull/56 and it worked out beautifully.

PR #100, opened 2016-12-29
Fix initialization alexa-app on a request without context

### LINKED ISSUE #96: Missing support for standard requests
As far as I can tell from code and from testing, this <a href="https://github.com/matt-kruse/alexa-app/commit/e2b8c6c13fad9bec2c0ad5dbcaf899791c12a02d">commit</a> has unintentionally dropped support for standard requests. I was testing upgrading to latest in master today to try out the new AudioPlayer features and I found that none of my standard launch or intent requests would go through at all.  Started debugging and found on line 155 of index.js that "this.data.context" is undefined and produces a catch error that breaks execution.  The JSON structure for AudioPlayer requests includes the context attribute and as such is supported by the commit, but the original support for standard requests which include the session attribute was completely dropped.  Not sure how this was overlooked, but should be fixed.  I am running an alexa-app-server instance to debug locally.  If there is some obvious fix to this that I overlooked than it should probably be documented, but at the least the new commit should support legacy code.

fremail:
Fixes #96 

Try to get `userId`, `applicationId` and `context` object from `json.context` or `json.request.context`. 
In case of absense `context` in the request json try to get data from `json.session`.
If the request is w/o session, `userId`, `applicationId` and `context` will be null.

Also I removed `context` object from test requests which I had added before.

### DIFF
--- a/index.js
+++ b/index.js
@@ -183,36 +183,6 @@ alexa.request = function(json) {
       return null;
     }
   };
-  this.userId = this.data.context.System.user.userId;
-  this.applicationId = this.data.context.System.application.applicationId;
-  this.context = function () {
-    try {
-      return this.data.request.context = {
-        "System": {
-          "application": {
-            "applicationId": this.applicationId
-          },
-          "user": {
-            "userId": this.data.session.user.userId,
-            "accessToken": this.data.session.accessToken
-          },
-          "device": {
-            "supportedInterfaces": {
-              "AudioPlayer": {}
-            }
-          }
-        },
-        "AudioPlayer": {
-          "token": this.sessionId,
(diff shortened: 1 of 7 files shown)


trayburn:
Respectfully, this is going the wrong way.  `context` is required on all requests and as such we shouldn't be fixing the code to handle the lack of its presence, instead we should be ensuring that it is always included in the tester generated code.

Citation:
https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#context-object

dblock:
What @trayburn is saying is that #101 is the right fix for this? I would like to see some test coverage there if everybody agrees?

fremail:
I agree with you. The #101 is the right one

PR #101, opened 2016-12-30
Context should not be secondary to session

trayburn:
According to Amazon's documentation, `context` is sent on every request (including AudioPlayer events) but `session` is only sent on standard requests.  As such we should not be clobbering the submitted context with information from context.  This is particularly dangerous when you also are clobbering `token` with `sessionId`, which makes it impossible to track which stream is currently playing on the client.

Citation:
https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#context-object
https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#session-object

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [#90](https://github.com/matt-kruse/alexa-app/pull/90): Added [Danger](http://danger.systems), PR linter - [@dblock](https://github.com/dblock).
 * [#91](https://github.com/matt-kruse/alexa-app/pull/91): Fixed "Cannot read property 'new' of undefined" init error - [@fremail](https://github.com/fremail).
 * [#88](https://github.com/matt-kruse/alexa-app/pull/88), [#92](https://github.com/matt-kruse/alexa-app/pull/92): AudioPlayer functionality - [@wschaeferiii](https://github.com/wschaeferiii) and [@fremail](https://github.com/fremail).
+* [#101](https://github.com/matt-kruse/alexa-app/pull/101): Expect request `context` to always be present and not modify that data. [@trayburn](https://github.com/trayburn)
 * Your contribution here.
 
 ### 2.3.4 (May 23, 2016)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 * [#90](https://github.com/matt-kruse/alexa-app/pull/90): Added [Danger](http://danger.systems), PR linter - [@dblock](https://github.com/dblock).
 * [#91](https://github.com/matt-kruse/alexa-app/pull/91): Fixed "Cannot read property 'new' of undefined" init error - [@fremail](https://github.com/fremail).
 * [#88](https://github.com/matt-kruse/alexa-app/pull/88), [#92](https://github.com/matt-kruse/alexa-app/pull/92): AudioPlayer functionality - [@wschaeferiii](https://github.com/wschaeferiii) and [@fremail](https://github.com/fremail).
-* [#101](https://github.com/matt-kruse/alexa-app/pull/101): Expect request `context` to always be present and not modify that data. [@trayburn](https://github.com/trayburn)
+* [#101](https://github.com/matt-kruse/alexa-app/pull/101): Added `request.context` - [@trayburn](https://github.com/trayburn).
 * Your contribution here.
 
 ### 2.3.4 (May 23, 2016)
--- a/index.js
+++ b/index.js
@@ -183,9 +183,16 @@ alexa.request = function(json) {
       return null;
     }
   };
-  this.userId = this.data.context.System.user.userId;
-  this.applicationId = this.data.context.System.application.applicationId;
-  this.context = this.data.context;
+
+  this.userId = null;
+  this.applicationId = null;
+  this.context = null;
+
+  if (this.data.context) {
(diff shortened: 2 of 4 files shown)


dblock:
I can merge, but what worries me here is that no tests broke. Maybe you want to reproduce this scenario in a test so that we never (un)break this again?

trayburn:
I agree that was suprising to me as well, I've added a test that ensures that `new alexa.request(json)` has a `context` property which hasn't changed at least the `applicationId`.

fremail:
One moment guys!

I just tested my skill from Amazon Dev Testing Tool and faced with a request without `context`. So this PR is right, but doesn't provide ability to test on testing tools :(

I did this request:
```
{
  "version": "1.0",
  "session": {
    "new": true,
    "sessionId": "amzn1.echo-api.session.0000000-0000-0000-0000-00000000000",
    "application": {
      "applicationId": "amzn1.ask.skill.[my_app_id]"
    },
    "attributes": {},
    "user": {
      "userId": "amzn1.account.AM3B00000000000000000000000"
    }
  },
  "context": {
    "System": {
      "application": {
        "applicationId": "amzn1.ask.skill.[my_app_id]"
      },
      "user": {
        "userId": "amzn1.account.AM3B00000000000000000000000"
      },
      "device": {
        "supportedInterfaces": {
          "AudioPlayer": {}
        }
      }
    },
    "AudioPlayer": {
      "offsetInMilliseconds": 0,
      "playerActivity": "IDLE"
    }
  },
  "request": {
    "type": "LaunchRequest",
    "requestId": "amzn1.echo-api.request.0000000-0000-0000-0000-00000000000",
    "timestamp": "2015-05-13T12:34:56Z",
    "locale": "string"
  }
}
```

But `request` (a first argument) in the `app.pre` was w/o `context`:
```
{ data: { version: '1.0', session: { new: true, sessionId: 'amzn1.echo-api.session.0000000-0000-0000-0000-00000000000', application: [Object], attributes: {}, user: [Object] }, request: { type: 'LaunchRequest', requestId: 'amzn1.echo-api.request.0000000-0000-0000-0000-00000000000', timestamp: '2016-12-30T07:34:18Z' } }, slot: [Function], type: [Function], hasSession: [Function], getSession: [Function], userId: 'amzn1.account.AM3B00000000000000000000000', applicationId: 'amzn1.ask.skill.[my_app_id]', context: [Function], sessionDetails: { new: true, sessionId: 'amzn1.echo-api.session.0000000-0000-0000-0000-00000000000', userId: 'amzn1.account.AM3B00000000000000000000000', accessToken: null, attributes: {}, application: { applicationId: 'amzn1.ask.skill.[my_app_id]' } }, sessionId: 'amzn1.echo-api.session.0000000-0000-0000-0000-00000000000', sessionAttributes: {}, isSessionNew: true, session: [Function] }
```

My skill is hosted on Lambda if it makes sense.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [#90](https://github.com/matt-kruse/alexa-app/pull/90): Added [Danger](http://danger.systems), PR linter - [@dblock](https://github.com/dblock).
 * [#91](https://github.com/matt-kruse/alexa-app/pull/91): Fixed "Cannot read property 'new' of undefined" init error - [@fremail](https://github.com/fremail).
 * [#88](https://github.com/matt-kruse/alexa-app/pull/88), [#92](https://github.com/matt-kruse/alexa-app/pull/92): AudioPlayer functionality - [@wschaeferiii](https://github.com/wschaeferiii) and [@fremail](https://github.com/fremail).
+* [#101](https://github.com/matt-kruse/alexa-app/pull/101): Expect request `context` to always be present and not modify that data. [@trayburn](https://github.com/trayburn)
```
dblock: Sorry to be nitpicky. Lets make this the same format as everything? That is `- [@trayburn](https://github.com/trayburn).`

dblock: Also, this didn't exist in 2.3.4 (double check please), so the change should be something like "Added `request.context`", right?

dblock:
@fremail @trayburn In theory I don't think we should be constructing a fake context for a testing scenario. What do you think we should do?

dblock:
@fremail with the code as in this PR, what's the first call/thing to break in your app?

fremail:
@dblock I check `applicationId` and if it's wrong (or null), the app doesn't work. That's for security reasons.

I agree, making fake context is not a good concept. I think we need to warn users that they may catch such error on testing.

dblock:
Agreed @fremail, I think `applicationId` should return `null` when not supplied and the app could only make that check in "production". Maybe extend the example in README and add an explanation?

@trayburn I'll merge after you address my comments above, thanks for bringing this up and fixing it!

index.js
```diff
@@ -185,34 +185,7 @@ alexa.request = function(json) {
   };
   this.userId = this.data.context.System.user.userId;
```
fremail: @dblock it fails there if this.data doesn't include context.

dblock: I suggest fixing and adding an `if`, `userId` should just be null here.

trayburn:
I believe that this update to the PR addresses all of the comments that have been brought up here.  If I missed something please let me know.

dblock:
Perfect, merged.

Issue #98, opened 2016-12-28
req.session throws "key not found on session attribute" 


cpup22:
I am trying to get a value from the session that isn't in the session:

req.session("keynotthere") and it throws 
key not found on session attributes: machine TypeError: Cannot read property 'keynotthere' of undefined

Which is fine but it won't continue to execute the logic in my code. I tried catching with a try catch but that doesn't work. I also tried printing out some messaging in the alexaapp.error method but it doesn't seem to get there.  I noticed there was another issue opened about try/catch handling, not sure if this is related.  

I mostly just want to check to see if the value exists in the session, if it does I handle, if it doesn't that's okay too.

dblock:
This is on HEAD or in the released version?

Find out what that `undefined` is that throws the error?

The version on HEAD has made significant strides around session objects, has methods like `hasSession`, `session.get`, etc., and we should add a `hasKey` method to the session object as well to solve your issue permanently.

cpup22:
released version I've been pulling/installing using npm.

dblock:
@cpup22 It would be useful to try against HEAD, we've made a lot of code changes here. Maybe this is already fixed.

dblock:
@cpup22 would like this confirmed as fixed/working on HEAD before releasing 2.4.0 (#103), will you be able to?

dblock:
Thx @trayburn for a test, this is fixed on HEAD, closed.

dblock:
@cpup22 Not yet, see https://github.com/matt-kruse/alexa-app/issues/103 there're other (potential) problems.

dblock:
...but do try HEAD please

PR #109, opened 2017-01-02
Fix #107: surface errors correctly, return a promise, no expectations.

### LINKED ISSUE #107: Build is incorrectly reporting success
```
Unhandled rejection AssertionError: expected 'INITIAL TEST VALUE WHICH SHOULD BE OVERRIDDEN' to equal undefined
    at assertEqual (/home/travis/build/matt-kruse/alexa-app/node_modules/chai/lib/chai/core/assertions.js:487:12)
    at ctx.(anonymous function) (/home/travis/build/matt-kruse/alexa-app/node_modules/chai/lib/chai/utils/addMethod.js:41:25)
    at doAsserterAsyncAndAddThen (/home/travis/build/matt-kruse/alexa-app/node_modules/chai-as-promised/lib/chai-as-promised.js:296:33)
    at null.<anonymous> (/home/travis/build/matt-kruse/alexa-app/node_modules/chai-as-promised/lib/chai-as-promised.js:255:21)
    at ctx.(anonymous function) [as equal] (/home/travis/build/matt-kruse/alexa-app/node_modules/chai/lib/chai/utils/overwriteMethod.js:49:33)
    at /home/travis/build/matt-kruse/alexa-app/test/test_alexa_app_session.js:229:55
    at tryCatcher (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/util.js:26:23)
    at Promise._settlePromiseFromHandler (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/promise.js:510:31)
    at Promise._settlePromiseAt (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/promise.js:584:18)
    at Promise._settlePromiseAtPostResolution (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/promise.js:248:10)
    at Async._drainQueue (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/async.js:128:12)
    at Async._drainQueues (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/async.js:133:10)
    at Immediate.Async.drainQueues [as _onImmediate] (/home/travis/build/matt-kruse/alexa-app/node_modules/bluebird/js/main/async.js:15:14)
    at tryOnImmediate (timers.js:534:15)
    at processImmediate [as _immediateCallback] (timers.js:514:5)
```

https://travis-ci.org/matt-kruse/alexa-app/builds/187971131

dblock:


test/test_alexa_app_session.js
```diff
@@ -224,10 +224,9 @@ describe("Alexa", function() {
           };
         });
 
-        it("session.get(key) should not throw if attribute is not present", function() {
+        it("session.get(key) should throw if attribute is not present", function() {
```
dblock: This now throws because we don't have a session (`{}.get(...)`). Is that correct?

```js
returnedAttributeValue = req.getSession().get("AttributeWhichDoesNotExist");
```

trayburn: Ok, back from my errands now, sorry for the delay.  I do not believe this is correct, I believe the requested behavior in #98 was that we return undefined.  The problem I believe was simply one of how I wrote the test.  I wanted two asserts, one that the `app.request()` call did not throw, the other that the `session.get()` call returned undefined, but by using a `.then()` I mucked with the `Promise`.

Summing up I believe the desired state is:
1. `session.get("unknown_key")` should return undefined
1. The `app.request(data)` call should not throw just because of that missing key

dblock: But this is in the case of `NO_SESSION`, what am I missing?

dblock:
@trayburn would appreciate a code review of this whole thing here

test/test_alexa_app_session.js
dblock: I believe (2) is checked with my change, but I could be wrong ;)

trayburn:
Review complete, so it gets my +1

dblock:
@trayburn what about that `NO_SESSION` thing above?

PR #118, opened 2017-01-11
Prevent updating session attributes directly

### LINKED ISSUE #117: Session object is updated directly and set in response
Currently session attributes are set when the response is prepared. When you get the session either with `req.session('key')` or `req.getSession` you get the session object back and modifying object properties on it will update the session in the response, e.g.

    const sesh = req.getSession();
    const objProp = sesh.get('objProp');
    objProp.existingProp = "new value";

This will set the `session.objProp.existingProp` to `"new value"` in the response even if it was not before.

My suggestion to fix this is to have `session.get` return a deep copy of the value. You will have to explicitly `session.set` in order to update the value.

ajcrites:
This change prevents users from updating the session attributes on the session object directly when working with the `session.getAttributes` and `session.get` methods.

Fixes #117

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * [#119](https://github.com/alexa-js/alexa-app/pull/119): Moved to the [alexa-js organization](https://github.com/alexa-js) - [@dblock](https://github.com/dblock).
 * Your contribution here.
-* [#118](https://github.com/matt-kruse/alexa-app/pull/118): [#117](https://github.com/matt-kruse/alexa-app/issues/117) Prevent updating session attributes directly - [@ajcrites](https://github.com/ajcrites).
+* [#118](https://github.com/matt-kruse/alexa-app/pull/118), [#117](https://github.com/matt-kruse/alexa-app/issues/117): Prevent updating session attributes directly - [@ajcrites](https://github.com/ajcrites).
 
 ### 2.4.0 (January 5, 2017)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 2.4.1 (Next)
 
 * Your contribution here.
+* [#118](https://github.com/matt-kruse/alexa-app/pull/118): [#117](https://github.com/matt-kruse/alexa-app/issues/117) Prevent updating session attributes directly - [@ajcrites](https://github.com/ajcrites).
```
dblock: Make that `:` a `,` and add a colon before `Prevent ...`.

index.js
```diff
@@ -262,7 +264,9 @@ alexa.session = function(session) {
   }
   this.getAttributes = function() {
     // do some stuff with session data
```
dblock: This feels like a useless comment that was left in there, feel free to better document these methods. Thanks!

ajcrites:
I'm not quite sure what you mean by the lower level session test and the `pre` dance. If you can clarify I can include these tests as well.

dblock:
If you don't mind rebasing this, please? See conflicting file above.

By "lower level session test" I mean that it seems that we're describing the behavior of a _session_, but are testing the behavior of an _app_. Session object is a thing, so I would imagine we would have tests for the session object itself, without the app. But that's nice to have, not required for me to merge this PR.

Thanks for your excellent work!

dblock:
Merged. I made version changes clearer in UPGRADING in https://github.com/alexa-js/alexa-app/commit/49cea22c559b310eae402ae868b57f67d9d1b7e1 too if you want to re-review that.

PR #120, opened 2017-01-12
Modified info about alexa-verifier module

tejashah88:
Small changes with information about alexa-verifier.
Removing reference to example about integrating with express since it's moved to a separate module (alexa-verifier-middleware).


dblock:
Merged, thanks.

PR #126, opened 2017-01-17
Rename lamda.js to lambda.js

tejashah88:
renamed `lamda.js` to `lambda.js` for consistency with README.


dblock:
Oh nice catch. Thanks.

PR #139, opened 2017-01-31
Mocha Testing changes

tejashah88:
* Added more node versions to run mocha tests on Travis-CI
* Changed arrow functions to normal functions (see https://mochajs.org/#arrow-functions for more info)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ language: node_js
 sudo: false
 
 node_js:
-- '0.10'
 - '0.12'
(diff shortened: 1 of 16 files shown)


dblock:
So the change is really about running tests on multiple versions of node, right? How about rewriting changelog to say "Compatibility with Node.js 0.12-7" or something like that? Fat arrows in tests is really not worth calling out :) I'll merge, if you feel like it you can amend changelog in a future PR.

PR #142, opened 2017-01-31
Fixed incorrect reference to PR

tejashah88:


tejashah88:
I don't know if it's just me, or is the coveralls bot reporting multiple times, just because we are testing with multiple versions of node on travis?

dblock:
Yes @tejashah88, it's annoying. You can make it run once for a specific node version and it will be better.

PR #144, opened 2017-02-01
refactoring the express() function

mreinstein:
people are having a lot of difficulty getting their middlewares defined in the right order. Especially on express apps that serve other traffic besides alexa skills.

This PR modifies the format of the call, and simplifies the usage.

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,9 +556,9 @@ app.launch(function(request,response) {
 });
 
 // ALWAYS setup the alexa app and attach it to express before anything else.
-app.express(express_app, express.Router(), "/echo/");
+app.express(express_app, express.Router());
 
-// now POST calls to /echo in express will be handled by the app.request() function
+// now POST calls to /sample in express will be handled by the app.request() function
 // GET calls will not be handled
 
 // from here on, you can setup any other express routes or middlewares as normal
--- a/README.md
+++ b/README.md
@@ -623,7 +623,6 @@ All named apps can be found in the `alexa.apps` object, keyed by name. The value
 
 Generally, an alexa-app module can be used inside a stand-alone Node.js app, within an HTTPS server or within an AWS Lambda function. The library only cares about JSON in and JSON out. It is agnostic about the environment that is using it, but it provides some convenience methods to hook into common environments.
 
-If you don't use AWS Lambda and host an Alexa skill on your own webserver, you will need to validate that requests come from Alexa. This validation is *not* provided by this module. For more details on how to handle alexa request validation, look at [alexa-verifier](https://github.com/alexa-js/alexa-verifier) which provides the necessary code.
 
 ### Development
 
--- a/README.md
+++ b/README.md
@@ -555,15 +555,14 @@ app.launch(function(request,response) {
   response.say("Hello World");
 });
 
-// this call defines a post() and optionally a get() handler in express, mapped to the alexa-app
-// express_app: the express app instance to map to
-//        path: the path prefix to map to
-// enableDebug: when false, don't map a GET handler, default is true
-//              debugging GET requests call express' render() method using 'test'
-app.express(express_app, "/echo/", false);
+// ALWAYS setup the alexa app and attach it to express before anything else.
+app.express(express_app, express.Router(), "/echo/");
 
-// now POST calls to /echo/sample in express will be handled by the app.request() function
+// now POST calls to /echo in express will be handled by the app.request() function
 // GET calls will not be handled
+
+// from here on, you can setup any other express routes or middlewares as normal
+
 ```
 
 ## Customizing Default Error Messages
--- a/README.md
+++ b/README.md
@@ -556,7 +556,7 @@ app.launch(function(request,response) {
 });
 
 // ALWAYS setup the alexa app and attach it to express before anything else.
-app.express(express_app, express.Router());
+app.express({ expressApp: express_app, router: express.Router() });
 
 // now POST calls to /sample in express will be handled by the app.request() function
 // GET calls will not be handled
--- a/example/express.js
+++ b/example/express.js
@@ -13,14 +13,14 @@ alexaApp.express({
   expressApp: app,
   router: express.Router(),
 
-  // checkCert verifies requests come from amazon alexa. Must be enabled for production.
-  // you can disable this if you're running a dev environment and want to POST things
-  // to test behavior. enabled by default.
+  // verifies requests come from amazon alexa. Must be enabled for production.
+  // You can disable this if you're running a dev environment and want to POST
+  // things to test behavior. enabled by default.
   checkCert: false,
 
-  // enableDebug sets up a GET route when set to true. This is handy for testing in 
+  // sets up a GET route when set to true. This is handy for testing in 
   // development, but not recommended for production. disabled by default
-  enableDebug: true
+  debug: true
 });
 
 
--- a/example/express.js
+++ b/example/express.js
@@ -17,9 +17,13 @@ var checkCert = false
 // enableDebug sets up a GET route when set to true. This is handy for testing in 
 // development, but not recommended for production. disabled by default
 var enableDebug = true
-alexaApp.express(app, express.Router(), "/echo/", checkCert, enableDebug);
 
-// here you can setup any other express routes or middlewares as normal
+alexaApp.express(app, express.Router(), checkCert, enableDebug);
+
+// now POST calls to /test in express will be handled by the app.request() function
+
+
+// from here on you can setup any other express routes or middlewares as normal
 app.use(bodyParser.urlencoded({ extended: true }));
 app.use(bodyParser.json());
 app.set("view engine", "ejs");
@@ -43,4 +47,4 @@ alexaApp.intent("nameIntent", {
 );
 
 app.listen(PORT);
-console.log("Listening on port " + PORT + ", try http://localhost:" + PORT + "/echo/test");
+console.log("Listening on port " + PORT + ", try http://localhost:" + PORT + "/test");
--- a/example/express.js
+++ b/example/express.js
@@ -9,16 +9,20 @@ var app = express();
 // ALWAYS setup the alexa app and attach it to express before anything else.
 var alexaApp = new alexa.app("test");
 
-// checkCert verifies requests come from amazon alexa. Must be enabled for production.
-// you can disable this if you're running a dev environment and want to POST things
-// to test behavior. enabled by default.
-var checkCert = false
+alexaApp.express({
+  expressApp: app,
+  router: express.Router(),
 
-// enableDebug sets up a GET route when set to true. This is handy for testing in 
-// development, but not recommended for production. disabled by default
-var enableDebug = true
+  // checkCert verifies requests come from amazon alexa. Must be enabled for production.
+  // you can disable this if you're running a dev environment and want to POST things
+  // to test behavior. enabled by default.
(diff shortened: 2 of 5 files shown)


mreinstein:
DON'T MERGE ME YET. Still a work in progress. Needs review and testing. this is a breaking change which will force us to increment the major version.

mreinstein:
@tejashah88 @dblock this is ready for review. Several breaking API changes:

* changed function prototype to: 

```javascript
this.express = function(express, router, certCheck, enableDebug) {
```
invoking this function now requires an express router to be passed in as the 2nd parameter. This is to ensure the middlewares that run in `alexa-app` are totally separate from the rest of the express instance.

* changed `enableDebug` to be false by default. I think default settings should be safe and assume production. dev settings should be opt-in.

* this module now internally calls the `alexa-verifier-middleware`

* the way the paths attach is slightly different. The old way:

```javascript
var app = new alexa.app("sample");
app.express(express_app, "/echo/", false);
// express is now routing POST /echo/sample to the alexa-app instance
```

new way:

```
var app = new alexa.app("sample");
app.express(express_app, express.Router());

// express is now routing POST /sample to the alexa-app instance
```

looking for feedback on this. I've set up a small alexa test skill on my end and this seems to work correctly. 

Also, If we are going to break the API and go to `alexa-app@3.0.0` we might as well find any other breaking changes and lump them in to. For example: maybe instead of `function(express, router, certCheck, enableDebug)` we pass an options object.

test/test_alexa_integration_express.js
```diff
@@ -73,7 +73,7 @@ describe("Alexa", function() {
 
     context("#express with debug set to true", function() {
       beforeEach(function() {
-        testApp.express(app, '/', true)
+        testApp.express(app, , express.Router(), false, true)
```
tejashah88: You should probably fix that extraneous comma.

tejashah88:
Looking good so far. I do agree that an `options` object format would make more sense. All we need to do is update the CHANGELOG accordingly and we should be set. But let's see what @dblock has to say.

mreinstein:
another possible breaking api change: right now we have:
```javascript
var app = new alexa.app(appname, endpoint)
```

but `endpoint` is only used in the context of hooking it up to express. We might want to eliminate it from the constructor and include it as an optional argument that gets passed into  `app.express(options)`

rickwargo:
@mreinstein I was struggling yesterday with something similar. I working on a nice UI to do the testing (have it) and noticed that a POST from my UI was being swallowed. I believe in the express.use in self.start, next() needs to be called if the method is POST. This worked in one scenario, but I have not done much testing. Plan to look into that today - and this! Thank you.

dblock:
This looks overall good. I definitely think `app.express(options)` is the better way to go, it's forward compatible, we can keep adding options. 

UPGRADING.md needs an explanation of how to upgrade from 2.x to 3.x.

example/express.js
```diff
@@ -2,14 +2,33 @@ var express = require("express");
...
+var alexaApp = new alexa.app("test");
+
+// checkCert verifies requests come from amazon alexa. Must be enabled for production.
+// you can disable this if you're running a dev environment and want to POST things
+// to test behavior. enabled by default.
+var checkCert = false
+
+// enableDebug sets up a GET route when set to true. This is handy for testing in 
+// development, but not recommended for production. disabled by default
+var enableDebug = true
```
dblock: This will read so much better with

```
var options = {
 checkCert: false,
 enableDebug: true
}
```

index.js
```diff
@@ -544,10 +545,34 @@ alexa.app = function(name, endpoint) {
...
-    var endpoint = (path || "/") + (self.endpoint || self.name);
-    express.post(endpoint, function(req, res) {
+  // A utility method to bootstrap alexa endpoints into an express router
+  //
+  // @param object express the express app to attach to
+  // @param router express router instance to attach to the express app
+  // @param bool certCheck when true, applies alexa certificate checking (default true)
+  // @param bool enableDebug when true, sets up the route to handle GET requests (default false)
+  // @return ? TODO: not sure what this returns
+  this.express = function(express, router, certCheck, enableDebug) {
```
dblock: This could probably raise an error if the second parameter is not a router (eg. a string), so that someone upgrading can "see" it.

example/express.js
```diff
@@ -2,14 +2,33 @@ var express = require("express");
...
 
+alexaApp.express(app, express.Router(), checkCert, enableDebug);
```
dblock: We call this `app` here, but maybe to be consistent it should be called `express`?

mreinstein:
@rickwargo thanks for the feedback. I'm not sure I fully understand what you mean, can you please elaborate?

rickwargo:
@mreinstein haven't had a chance to fully look into it yet. My problem is with the `alexa-app-server`. Turns out that a POST 9or other non-GET method) is not propagated down the middleware stack in debugging due to the following code:
```
// Configure GET requests to run a debugger UI
if (false!==config.debug) {
	self.express.get(endpoint,function(req,res) {
		if (typeof req.param('schema')!="undefined") {
			res.set('Content-Type', 'text/plain').send(app.schema());
		}
		else if (typeof req.param('utterances')!="undefined") {
			res.set('Content-Type', 'text/plain').send(app.utterances());
		}
		else {
			res.render('test',{"app":app,"schema":app.schema(),"customSlotTypes":(app.customSlotTypes?app.customSlotTypes():""),"utterances":app.utterances(),"intents":app.intents});
		}
	});
}
```
It's the render of test that causes it. A temporary work-around (what I am currently employing but need to investigate the real cause) is to call `next()` in the "rawBody" middleware baked into `self.start`. I've added:
```
if (false===config.debug || 'GET'!==req.method)
	next();
```
to the end of this function in self.start as a temporary measure:
```
//We need the rawBody for request verification
self.express.use(function(req, res, next)
{
```
My problem surfaced as I am creating a UI around testing multiple apps. When I perform the POST to the alexa-app-server, it was swallowed by one of the middlewares (the one above) and my code was not executed (left in a PENDING state on the client browser).

Does that help?

mreinstein:
@rickwargo it sounds like this is an `alexa-app-server`issue. I haven't had a chance to fix it. In fact I don't even think it'll pass alexa certification process as-is. Might want to repost your comment in that repo.

I'm hoping to do a refactor on that this afternoon.

mreinstein:
changes finalized. Should be good-to-go. @tejashah88 @rickwargo @dblock feel free to review again if desired.

README.md
```diff
@@ -555,15 +555,14 @@ app.launch(function(request,response) {
...
-// this call defines a post() and optionally a get() handler in express, mapped to the alexa-app
-// express_app: the express app instance to map to
-//        path: the path prefix to map to
-// enableDebug: when false, don't map a GET handler, default is true
-//              debugging GET requests call express' render() method using 'test'
-app.express(express_app, "/echo/", false);
-
-// now POST calls to /echo/sample in express will be handled by the app.request() function
+// ALWAYS setup the alexa app and attach it to express before anything else.
+app.express({ expressApp: express_app, router: express.Router() });
```
dblock: I might be nitpicking, but why don't we just call this parameter `app`, otherwise we have to call `router`, `expressRouter`?

index.js
```diff
@@ -178,12 +180,11 @@ alexa.request = function(json) {
     }
   };
   this.type = function() {
-    try {
-      return this.data.request.type;
-    } catch (e) {
-      console.error("missing type", e);
-      return null;
+    if (!(this.data && this.data.request && this.data.request.type)) {
+      console.error("missing request type:", this.data);
```
dblock: Needs a space after `:`.

dblock:
This looks good with the renaming suggested above, thanks @mreinstein!

README.md
mreinstein: because app is too vague as it is :(  This looks less clear what is happening:

```javascript
app.express({ app: express_app, router: express.Router() });
```

`router` is short and it's distinct (express is the only entity that has a router construct in this context.)

dblock: Dumb question, why can't `router` be internal to the `express` method, I mean why do we have to pass it in, isn't it just `require("express").Router()`?

mreinstein: > why can't router be internal to the express method

Not a dumb question. I looked at this initially but then we'd be adding `express` as a dependency to `alexa-app`, and it might track differently from whatever express version someone is using in their code. (I don't want to get into the scenario where `alexa-app` ends up mounting stuff from 2 different versions of express.)

index.js
mreinstein: Unfortunately I have to put down these alexa PRs and get back to completing some other work for the time being. If someone else wants to pick this up and run with it, should be mostly there. Happy to answer questions. Or I can get back to this next week.

README.md
dblock: And I suppose you're against just requiring it without an explicit dependency? When you do a `require` it pulls it out of the parent application's space anyway.

index.js
dblock: IT's super minor, nbd.

dblock:
This is a step in the right direction, so I'll just merge.

README.md
mreinstein: `require()`ing a dependency that isn't in the application's `package.json` feels icky, even though it should work fine.

Another nice consequence of passing it in is you have more control over how you mount routes in the app. For example, in the still-open PR for we use it to pass an instance of a router instead of an express app:
https://github.com/mreinstein/alexa-app-server/blob/master/index.js#L100

dblock: I agree it feels icky. I feel like it would be OK to default it though and leave the option out of README examples, but I don't feel strongly about it at all.

mreinstein: actually, it would cause breakage for anyone that is using `alexa-app` without using `express`. (e.g., AWS Lambda contexts)

dblock: Not if we do that inside a function. It's actually perfectly fine to do this - in fact in some situations it's better. Node caches every module in memory, it doesn't reload them each time you require one.

mreinstein: I still wouldn't

mreinstein: while we're breaking old API, it might be nice to rename the function. perhaps something like:

```javascript
app.attachToExpress({ app: express_app, router: express.Router() });
```

instead, to be more clear what the function actually does

dblock: I like that. Maybe `mount`?

tejashah88: It would also make sense to rename the lambda function to `app.attachToLambda()` or something.

index.js
```diff
@@ -544,21 +544,48 @@ alexa.app = function(name, endpoint) {
...
       });
     });
-    if (typeof enableDebug != "boolean") {
-      enableDebug = true;
-    }
-    if (enableDebug) {
-      express.get(endpoint, function(req, res) {
+
+    if (options.debug) {
+      router.get("/", function(req, res) {
```
tejashah88: One issue that I just found while testing. When you do a GET request and the verifier middleware is enabled with strict header checking, it gives me a 401 error, which shouldn't happen at all.

One possible fix is that the alexa-verifier-middleware should check what request method is being sent and if it is a POST request, then process the request accordingly. [Here](https://expressjs.com/en/4x/api.html#req.method) is a way to access the request's method from expressjs.

mreinstein:
Good catch. I think we should install any routes that don't need checking before the Middleware is set up.

mreinstein:
Everything except debug route should have check Middleware enabled I believe so I think it's just a matter of moving this one route up

PR #145, opened 2017-02-02
debug route installed before verifier middleware.

mreinstein:
this enables debug mode to work properly.

### DIFF
--- a/index.js
+++ b/index.js
@@ -571,6 +571,16 @@ alexa.app = function(name) {
 
     options.expressApp.use(endpoint, router);
 
+    if (options.debug) {
+      router.get("/", function(req, res) {
+        res.render("test", {
+          "json": self,
+          "schema": self.schema(),
+          "utterances": self.utterances()
+        });
+      });
+    }
+
     if (options.checkCert) {
       options.router.use(verifierMiddleware({ strictHeaderCheck: true }));
     }
@@ -584,15 +594,6 @@ alexa.app = function(name) {
       });
     });
 
-    if (options.debug) {
-      router.get("/", function(req, res) {
-        res.render("test", {
-          "json": self,
-          "schema": self.schema(),
-          "utterances": self.utterances()
-        });
-      });
-    }
   };
 
   // Add the app to the global list of named apps


mreinstein:
this coveralls shit is incredibly obnoxious and verbose. Can it be turned down/off?

mreinstein:
@tejashah88 @dblock  do either of you know what this line is about? https://github.com/alexa-js/alexa-app/blob/master/index.js#L542

It seems like `.lamba()` and `.handler(event, context)` are both related to using lambda. Is one an old way which is deprecated/to be removed?

dblock:
Aha!

dblock:
I just turned off comments in coveralls settings, it's indeed super annoying. We'll just have to keep an eye on test coverage.

dblock:
I am using the lambda function in https://github.com/artsy/elderfield/blob/028b4d737c3344bd4bda41fe8daf35778840dbc8/functions/artsy/index.js#L281, but I suppose I could just use handler? I think this just existed before and we kept it. We can probably nuke it.

dblock:
We obviously need a test for this one, too.

dblock:
@mreinstein I was trying to write a test for this (#146) - how did this change fix anything? With `checkCert: true` if verification fails, you get a 401. If your verification succeeds you get the response from the debugger. Because the middleware hooks onto events why did it matter in which order these declarations came?

mreinstein:
the order matters because express evaluates middlewares in the order they are installed. The debug route will never be called by amazon, and should only be called in dev environments. Installing this middleware before the verifier check ensures that this verification doesn't happen on the GET route.

dblock:
Thanks, I didn't consider that one was GET and the other one POST only. I suppose https://github.com/alexa-js/alexa-verifier-middleware/pull/13 makes this part moot because it makes the verifier only care about POST, and is probably a better long term fix.

I put up https://github.com/alexa-js/alexa-app/pull/154, a test that would fail before the change here.

mreinstein:
> makes this part moot because it makes the verifier only care about POST, and is probably a better long term fix.

I disagree. By default, all of the alexa routes should have the verifier applied. We should be defaulting to production settings so that people have to opt into explicitly insecure behavior.

I also think moving the get route up is superior in that it introduces no new lines of code; it just orders them correctly.

dblock:
I think I agree, although all alexa routes are POST, no? I added [a comment](https://github.com/alexa-js/alexa-verifier-middleware/pull/13#issuecomment-277483372) to this effect.

tejashah88:
I'm curious to know why all alexa routes (GET and POST) should be verified. The way I see it, the GET requests are for debugging if the `alexa-app` is working in terms of if it's doing it's job. The verification should be for POST requests since that's the only requests that Amazon would be sending.

mreinstein:
> I'm curious to know why all alexa routes (GET and POST) should be verified

because by definition, being an alexa route means that it should be coming from amazon. the debug `GET` request is the only known exception to this. we should be effectively whitelisting the `get` route because this is safer behavior, and what should be done in production.  All defaults should assume production to prevent copy/paste monkeys from fucking up their own deploys.

tejashah88:
That does make sense. I'll remove the change for now.

PR #147, opened 2017-02-02
attachToExpress() , attachToLambda() function renaming

mreinstein:


dblock:
Sorry to be the process guy, but this one really needs a CHANGELOG entry and something in UPGRADING. It's so much easier to take the habit of documenting API changes while small changes are made than have to go through git history before a release and hope to uncover every single thing, don't you think? Lets make it easy on our users?

dblock:
Looks like @mreinstein deleted his fork, probably following https://github.com/alexa-js/alexa-app-server/pull/56. Closing this, renaming these functions is a good idea, but not required and can wait.

PR #150, opened 2017-02-03
Dump schema and utterances in debug mode and added preRequest and postRequest support.

dblock:
https://github.com/alexa-js/alexa-app-server/pull/56 attempts to call the express function from alexa-app instead of rewriting it. It seems logical to move some of the alexa-app-server debug functionality and pre and post request filtering further down.

### DIFF
--- a/test/test_alexa_integration_express.js
+++ b/test/test_alexa_integration_express.js
@@ -173,5 +173,36 @@ describe("Alexa", function() {
           });
       });
     });
+
+    context("#express with checkCert=true", function() {
+      beforeEach(function() {
+        testApp.express({
+          expressApp: app,
+          router: express.Router(),
+          checkCert: true,
+          debug: false
+        });
+      });
+
+      it("requires a cert header", function() {
+        return request(testServer)
+          .post('/testApp')
+          .expect(401).then(function(res) {
+            expect(res.body.status).to.equal("failure");
+            expect(res.body.reason).to.equal("The signaturecertchainurl HTTP request header is invalid!");
+          });
+      });
+
+      it("checks cert header", function() {
+        return request(testServer)
+          .post('/testApp')
+          .set('signaturecertchainurl', 'dummy')
+          .set('signature', 'dummy')
+          .expect(401).then(function(res) {
+            expect(res.body.status).to.equal("failure");
+            expect(res.body.reason).to.equal("signature is not base64 encoded");
+          });
+      });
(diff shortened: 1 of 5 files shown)


dblock:
@tejashah88, @rickwargo or @mreinstein Care to take a look at this?

rickwargo:
I was able to get the test.ejs to render and work, and also get schema and utterances with /alexa/hello_world?schema and /alexa/hello_world?utterances along with one of my apps.

dblock:
@rickwargo https://github.com/alexa-js/alexa-app/pull/150/commits/bfef3e2dec8f15b27cfa53e994282758e909c096

dblock:
@tejashah88 I gave you repo permissions, appreciate another pair of eyes/cr/merge.

test/test_alexa_integration_express.js
```diff
@@ -102,6 +113,24 @@ describe("Alexa", function() {
             expect(response.text).to.startWith('{"name":"testApp"')
           });
       });
+
+      it("returns debug schema", function() {
+        return request(testServer)
+          .get('/testApp?schema')
+          .expect(200).then(function(response) {
+            expect(response.headers['content-type']).to.equal('text/plain; charset=utf-8');
+            expect(response.text).to.include('"intent": "myIntent"')
```
tejashah88: It would be safer to compare response.text to the actual result. If the result is really big, you can load it from a file and compare from that. See [alexa-app-server/index.js#L85](https://github.com/alexa-js/alexa-app-server/blob/master/test/test-examples-server.js#L85) for an example.

test/test_alexa_integration_express.js
```diff
@@ -102,6 +113,24 @@ describe("Alexa", function() {
...
+          });
+      });
+
+      it("returns debug utterances", function() {
+        return request(testServer)
+          .get('/testApp?utterances')
+          .expect(200).then(function(response) {
+            expect(response.headers['content-type']).to.equal('text/plain; charset=utf-8');
+            expect(response.text).to.include('my name is bob')
```
tejashah88: Same as above. See [alexa-app-server/index.js#L93](https://github.com/alexa-js/alexa-app-server/blob/master/test/test-examples-server.js#L93) for an example.

dblock: Good point, even better updated to compare to the actual schema and utterances of the app.

PR #153, opened 2017-02-04
Document minimum node.js version, closes #143.

### LINKED ISSUE #143: Add section for which node versions are supported
If we plan to make any breaking changes that could potentially drop support for older versions of node, it's best to add a section in the README stating which versions **are** supported.

dblock:
Closes #143.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ A Node module to simplify development of Alexa apps (Skills) using Node.js.
 npm install alexa-app --save
 ```
 
+Node.js 0.12 or later is required.
+
 # Stable Release
 
 You're reading the documentation for the next release of alexa-app, which should be 2.4.1. Please see [CHANGELOG](CHANGELOG.md) and make sure to read [UPGRADING](UPGRADING.md) when upgrading from a previous version. The current stable release is [2.4.0](https://github.com/alexa-js/alexa-app/blob/v2.4.0/README.md).
--- a/package.json
+++ b/package.json
@@ -3,7 +3,10 @@
   "version": "2.5.0",
   "description": "A module to simplify creation of Alexa (Amazon Echo) apps (Skills) using Node.js",
   "main": "index.js",
-  "author": "Matt Kruse <github@mattkruse.com> (http://MattKruse.com)",
+  "author": "Matt Kruse <github@mattkruse.com> (http://mattkruse.com)",
+  "engines": {
+    "node": ">=0.12"
+  },
   "keywords": [
     "amazon",
     "echo",


tejashah88:
Looks good. I'll merge it!

mreinstein:
I think this would be better served by the `engines` field in `package.json`. It doesn't clutter the README (which is already so long as to be unreadable) and enables npm to use the field programmatically.

tejashah88:
IMO, it really doesn't clutter the README by that much, and it's visible so that anyone that just clicks on this module without scrolling too far down can see it.

tejashah88:
The real problem with the clutter is that most of it just talks about using the library, which should practically be moved to the wiki section.

dblock:
@mreinstein I did add an entry to package.json *and* a note in README. Did I miss something? 

One thing about package.json entry, its enforcement doesn't seem to do what you'd expect it to do, when you check out this code for example and run `npm install` with 0.10, it doesn't complain. I think it does when you include alexa-app into something else.

dblock:
Speaking on README length (and I didn't write much of it), other libraries like [grape](https://github.com/ruby-grape/grape) have the longest README on planet earth, and multiple people praise it for being the best written thing ever. So YMMV. The important thing is that everything is documented. 

With a wiki it makes it hard to document something _together_ with code in a single pull request, so my vote would be not to do that.

tejashah88:
That might be because there is a table of contents, otherwise people would hate to scroll through all of that.

mreinstein:
> other libraries like grape have the longest README on planet earth, and multiple people praise it

This is anecdotal. I never suggested any kind of universal law about how long "good" READMEs are. 
All I'm saying is from the perspective of someone trying to use this, most don't want a tomb of stuff to sift through. A short thing to read that gets you up and running, with a pointer to more details as needed. 

If you can't shorten this at least put the stuff relevant to getting started at the top (what is this module, how do you start using it.) There's all this cruft at the top about stable releases, minimum node version, etc. most people just casually trying or using this don't care about that stuff.

dblock:
It sounds like you have a good handle on what you'd like that README to be @mreinstein, so you're more than welcome to give it a shot as always.

mreinstein:
yeah I'm just really slammed right now with contracting work. Maybe later tonight or tomorrow I'll separate the chaff from the wheat

dblock:
My goal for now is to focus on getting a release of alexa-app and alexa-app-server that work well together (as you have suggested before @mreinstein). Will leave READMEs for later.

mreinstein:
sounds good

PR #156, opened 2017-02-05
Don't run danger, linter and coverage on all node versions.

dblock:


.travis.yml
```diff
@@ -1,39 +1,32 @@
----
 language: node_js
 
 sudo: false
 
+matrix:
+  include:
+    - node_js: 0.12
```
tejashah88: Wouldn't it make sense to use the latest version of node with the other scripts?

dblock: It's fine too, but I don't think it matters much as long as they can run. Different versions of node don't produce different outputs from these hopefully.

PR #159, opened 2017-02-05
Fix #157: SessionEndedRequest hangs without a defined sessionEndedFunc.

### LINKED ISSUE #157: POST of SessionEndedRequest hangs if no sessionEndedFunc
Unless a sessionEndedFunc is defined, a POST will remain in the PENDing state on the client. It works correctly when defined in the Alexa app. I believe this is because response.send() is not called in the not defined scenario.

dblock:


dblock:
Thanks @rickwargo, merged.

Issue #134, opened 2017-01-27
Deprecate returning false and friends in favor of promises


dblock:
The change in https://github.com/alexa-js/alexa-app/pull/133 was backwards compatible, but we may want a major version change that supports only promises and just using:

```
Promise.resolve(handlerResult).then(res.send).catch(handleError);
```

We can also remove the `return false` functionality and make either returning a Promise or calling the callback required in order to send the response (or just call response.send() explicitly).

ajcrites:
@dblock do we still want to support callbacks as well or can we purely support promises? This will also apply to pre/post.

dblock:
I think a deprecation message should happen first, then a major version bump and only supporting promises with good documentation is 👍 for me.

ajcrites:
Do we want to have a separate issue for actually implementing the changes and keep this simply for adding a deprecation message and updating the documentation?

dblock:
We want to have some pull requests :)

dblock:
https://github.com/alexa-js/alexa-app/pull/167

ajcrites:
Can this issue be closed?

dblock:
Closed via https://github.com/alexa-js/alexa-app/pull/167. 

Also @ajcrites you have committer access don't you. Just do what you think is right and we can argue about it later ;)

PR #171, opened 2017-02-13
Updating library for asynchronous support through promises

### LINKED ISSUE #22: is it possible to do asynchronous things in pre() ?
I want my `pre()` to able to make asynchronous calls before running the request. like this:

``` coffeescript
app.pre = (request, response, type) ->
  getAccessToken request.data.session.user.userId, (er, token) ->
    if er or (not token)
      response.fail 'Sorry, you must link your device first.'
    else
      # TODO: proceed to normal intent handler
```

Is there a way to do asynchronous things in pre?

ajcrites:
This completes the change for #134. Asynchronous support is no longer backwards compatible, and the `return false` and callbacks to handlers have been removed. Now, returning Promises from handlers is required for asynchronous support.

`pre` and `post` now both also have asynchronous support. The API functions identically as before as long as you use the Promise chain for asynchronicity. `response.send` and `response.fail` can force immediate success or failure. `.post` can still recover failures. The `response.send` and `response.fail` methods return promises that must be returned from inside the promise chain in order to continue it.

Also closes #22

### DIFF
--- a/index.js
+++ b/index.js
@@ -434,7 +434,15 @@ alexa.app = function(name) {
     })
     .catch(function(e) {
       if (typeof self.error == "function") {
-        return self.error(e, request, response);
+        // Default behavior of any error handler is to send a response
+        return Promise.resolve(self.error(e, request, response)).then(function() {
+            if (!response.resolved) {
+                response.resolved = true;
+                return response.send();
+            }
+            // propagate successful response if it's already been resolved
+            return response.response;
+        });
       } else if (typeof e == "string" && self.messages[e]) {
         if (!request.isAudioPlayer()) {
           response.say(self.messages[e]);
--- a/index.js
+++ b/index.js
@@ -353,145 +353,95 @@ alexa.app = function(name) {
     self.sessionEndedFunc = func;
   };
   this.request = function(request_json) {
-    return new Promise(function(resolve, reject) {
-      var request = new alexa.request(request_json);
-      var response = new alexa.response(request.getSession());
+    var request = new alexa.request(request_json);
+    var response = new alexa.response(request.getSession());
+    var postExecuted = false;
+    var requestType = request.type();
+    var prePromise = Promise.resolve();
 
-      // error handling when a request fails in any way
-      var handleError = function(e) {
-        if (typeof self.error == "function") {
-          self.error(e, request, response);
-          if (!response.resolved) {
-              response.send();
-          }
-        } else if (typeof e == "string" && self.messages[e]) {
-          if (!request.isAudioPlayer()) {
-            response.say(self.messages[e]);
-            response.send(e);
-          } else {
-            response.fail(self.messages[e]);
-          }
-        }
-        if (!response.resolved) {
-          if (e.message) {
-            response.fail("Unhandled exception: " + e.message + ".", e);
-          } else if (typeof e == "string") {
-            response.fail("Unhandled exception: " + e + ".", e);
-          } else {
-            response.fail("Unhandled exception.", e);
-          }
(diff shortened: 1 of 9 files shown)


dblock:
If @mreinstein would like to do a constructive code review here that'd be much appreciated. Removing non-promised versions to simplify the code in the medium term and fixing bug #22 is what we're trying to accomplish.

dblock:
I merged https://github.com/alexa-js/alexa-app/pull/167, @ajcrites you should rebase this.

dblock:
Also see #173.

index.js
```diff
@@ -348,140 +348,104 @@ alexa.app = function(name) {
...
+          return response.send(e);
+        } else {
+          return response.fail(self.messages[e]);
+        }
+      }
+      if (!response.resolved) {
+        if (e.message) {
+          return response.fail("Unhandled exception: " + e.message + ".", e);
+        } else {
+          return response.fail("Unhandled exception.", e);
```
rickwargo: `e` often contains a string and this should be part of the `.fail() call, especially if `e` is an indexed message (`self.messages[e]`).

index.js
```diff
@@ -348,140 +348,104 @@ alexa.app = function(name) {
...
+        }
+      }
+    })
+    .then(function () {
+      return response.send();
+    })
+    .catch(function(e) {
+      if (typeof self.error == "function") {
+        self.error(e, request, response);
+      } else if (typeof e == "string" && self.messages[e]) {
```
rickwargo: I think the default behavior should be to resolve by calling `response.send()`here as the user error handler, `app.error()` can still explicitly fail if desired.

index.js
```diff
@@ -348,140 +348,104 @@ alexa.app = function(name) {
...
+          throw "INVALID_REQUEST_TYPE";
+        }
+      }
+    })
+    .then(function () {
+      return response.send();
+    })
+    .catch(function(e) {
+      if (typeof self.error == "function") {
+        self.error(e, request, response);
```
rickwargo: I think the default behavior should be to resolve by calling `response.send()` here as the user error handler, `app.error()` can still explicitly fail if desired.

index.js
ajcrites: Do you mean something like `response.fail("Unhandled exception: " + e, e)`?

index.js
ajcrites: See comments on #182 -- I can change this to `return self.error` and update the documentation to say that error can be async.

index.js
rickwargo: Exactly!

index.js
rickwargo: I think it should next expect the error handler to return anything in release v3. Instead a call to resolve.send() if it has not been already resolved should suffice. I submitted PR #183 to handle.

ajcrites:
@dblock this is now up-to-date with current changes and makes `.error` support asynchronously and default to resolving. I think that this will work best, and I tried to document this functionality as well.

Anything else we need to do to proceed with this update? We can work off of a v4 branch and create an rc rather than using master if we want to spend some time with v3.

dblock:
Can you take a look at https://github.com/alexa-js/alexa-app/pull/186 and merge it if it makes sense? I think @tejashah88 is not around. Then make a 3.2.0 release (follow RELEASING) and I will merge this PR right behind it?

tejashah88:
@dblock Sorry for slight inactivity. #186 is just merged.

dblock:
What are we paying you for @tejashah88  ;)

dblock:
I'm merging this.

PR #174, opened 2017-02-13
Always enfore strict header checking [broken]

tejashah88:
Updated `alexa-verifier-middleware` to `0.2.1` from `0.1.9`, which removes the `strictHeaderCheck` option. Either the verification should be completely on, or completely off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,13 @@
 ## Changelog
 
-### 3.0.1 (Next)
+### 4.0.0 (Next)
+* Your contribution here.
+
+### 3.1.0 (February 13, 2017)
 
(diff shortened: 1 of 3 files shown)


dblock:
This forces all kinds of upgrades on people, like bluebird, alexa-utterances, etc., while the library is perfectly compatible with older verisons. I think you should revert package.json and only do the upgrade that is required by the interface change.

dblock:
Also build seems to have broken because of the danger upgrade.

dblock:
Still broken, revert package.json and do only the update of the minimum version of alexa-verifier-middleware.

test/test_alexa_integration_express.js
```diff
@@ -183,15 +183,6 @@ describe("Alexa", function() {
         });
       });
 
-      it("requires a cert header", function() {
```
dblock: What's the reason for removing this test?

tejashah88:
@dblock New PR: #178

PR #176, opened 2017-02-13
add flexibility for attaching alexa-app to express

bgjehu:
add flexibility for attaching alexa-app to either an express app or router

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 3.0.2 (February 13, 2017)
+* [#176](https://github.com/alexa-js/alexa-app/pull/176): Added flexibility to attach alexa-app to express app or router - [@bgjehu](https://github.com/bgjehu).
+
 ### 3.0.1 (Next)
 
 * [#162](https://github.com/alexa-js/alexa-app/issues/162): Fix: do not generate empty slots in schema - [@dblock](https://github.com/dblock).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 ## Changelog
 
 ### 4.0.0 (Next)
+* [#176](https://github.com/alexa-js/alexa-app/pull/176): Added flexibility to attach alexa-app to express app or router - [@bgjehu](https://github.com/bgjehu).
 * Your contribution here.
 
 ### 3.1.0 (February 13, 2017)
 
-* [#176](https://github.com/alexa-js/alexa-app/pull/176): Added flexibility to attach alexa-app to express app or router - [@bgjehu](https://github.com/bgjehu).
 * [#162](https://github.com/alexa-js/alexa-app/issues/162): Fix: do not generate empty slots in schema - [@dblock](https://github.com/dblock).
 * [#134](https://github.com/alexa-js/alexa-app/pull/134): Adding deprecation notices for plan to use Promises for async functionality - [ajcrites](https://github.com/ajcrites).
 
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ app.intent("number", {
 );
 
 // setup the alexa app and attach it to express before anything else
-app.express({ expressApp: express_app, router: express.Router() });
+app.express({ expressApp: express_app });
 
 // now POST calls to /sample in express will be handled by the app.request() function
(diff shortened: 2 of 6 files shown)


dblock:
Seems reasonable, but needs tests, README and CHANGELOG updates, please. And a passing build.

bgjehu:
@dblock fair, will do, marked WIP for now

bgjehu:
@dblock I updated README, CHANGELOG, version (3.1.0) for today release, test and express example. However, code is less so that coverage decrease. Is there work around? Or I have to do something?

dblock:
Code coverage is just there to tell you which scenario(s) aren't tested, but don't worry about it.

I thought I understood what this was, but maybe I don't. Can you explain to me like like 2 yo what's going on here? Note that this is particularly used in https://github.com/alexa-js/alexa-app-server/blob/master/index.js#L122 and I don't think it's right to specify one or the other since they can be different things. At the very least README is unclear about it.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog
 
 ### 4.0.0 (Next)
+* [#176](https://github.com/alexa-js/alexa-app/pull/176): Added flexibility to attach alexa-app to express app or router - [@bgjehu](https://github.com/bgjehu).
```
dblock: I think this should be clearer. What you're really doing is allowing `expressApp` or `router` to be optional, no?

bgjehu: so before the change, users (whoever use the module) need to pass in both express app and express router, and endpoint to specify where to attach the router to the app. But there is no need to pass in both. If users don't use express router, but just an express app. they should be just attach to the app and go. And if they use express router, they wouldn't want to do attaching router to app through alexa-app module. They would desire just attach the alexa-apo to an express router. And they could handle the attaching router to app themselves outside alexa-app scope (meaning not thru alexa.express() function)

bgjehu: @dblock ok, I see now. In https://github.com/alexa-js/alexa-app-server/blob/master/index.js#L122 it shows that the expressApp could be an instance of a router, not necessary an instance of an app. But the param naming is confusing then.
And what would be the reason to bind alexa-app to newly created router in root and bind it to something else thru alexa.express(). Wouldn't it be clearer to users that alexa.express() takes in an express router, bind alexa-app to the router to given endpoint instead of bind to a new router under root and bind the router to given endpoint?

dblock: First, I just want to say that I didn't write this code, the latest iteration was @mreinstein who should comment. Now that I am paying attention I can see what you're saying, but lets try to make some sense out of it. 

A router or an app are the same thing, a Router instance is a complete middleware and routing system (or mini-app as express calls it). This code mounts the alexa app inside the router (or app) given to it. But then we still need to attach the new router to the express app, don't we? Which is the first thing done in the code:

```
var endpoint = "/" + options.endpoint;
expressApp.use(endpoint, router);
```

dblock:
@bgjehu what happenned? ;)

bgjehu:
@dblock accidental operation, opening new one

bgjehu:
@dblock new one here https://github.com/alexa-js/alexa-app/pull/177

CHANGELOG.md
bgjehu: @dblock you probably have the answer now. But just to keep it as reference. Yes, the router still need to be attach to an express app. But that should not be in the scope of alexa-app nor alexa.express() function

dblock: Does this mean we're breaking alexa-app-server? Maybe we should raise an exception if both app and router were specified?

bgjehu: @dblock I don't know how you guys handle the release. If the changes is going in to 4.0.0, it shouldn't break the alexa-app-server for alexa-app-server's package.json is using alexa-app ^3.0.0
and it is good idea to throw exception if both specified! Thank you!

tejashah88: > Maybe we should raise an exception if both app and router were specified?

@dblock If that is to be the case, then maybe we should simplify it to one option, something between `expressApp` and `router`.

bgjehu: with this `var target = options.expressApp || options.router;` change
it would take expressApp to attach to when both are specified
and I am working on doing a warning instead of throw an exception

PR #177, opened 2017-02-13
adding flexibility for attaching alexa-app to express

bgjehu:
change alexa.express()
so that it takes either an express app or router and attach alexa-app to it under given endpoint.
leave the logic of binding router to app out of the scope of alexa-app/alexa.express()


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog
 
 ### 4.0.0 (Next)
+* [#176](https://github.com/alexa-js/alexa-app/pull/176): Added flexibility to attach alexa-app to express app or router - [@bgjehu](https://github.com/bgjehu).
```
dblock: Needs a better explanation that "flexibility", maybe _Express router is now optional_?

dblock:
This looks right to me, but I have this feeling that we're missing something, cc: @tejashah88.

@bgjehu Try this against alexa-app-server please, any failing tests?

The PR needs to be rebased, and see my minor comment on CHANGELOG above.

Please add a paragraph to UPGRADING wrt what one should do if they specified both express app and router before.

tejashah88:
Looks great so far, except for that `endpoint` does not have any normalization (i.e. `////test` is valid). 

@bgjehu I've submitted PR #180, which adds a bunch of utility functions, one of which includes a [function](https://github.com/alexa-js/alexa-app/pull/180/files#diff-14ef994fa167d355bdaf9b44add54b87R20) to normalize the final endpoint result.

bgjehu:
@dblock I will work on better explanation on CHANGELOG.md, work on testing with alexa-app-server and  updating UPGRADING.md
@tejashah88 I will add the usage of normalization

index.js
```diff
@@ -560,25 +560,20 @@ alexa.app = function(name) {
...
+    var defaultOptions = { endpoint: "/" + self.name, checkCert: true, debug: false };
 
     options = defaults(options, defaultOptions);
 
-    var endpoint = "/" + options.endpoint;
-    var router = options.router;
-
-    options.expressApp.use(endpoint, router);
+    // In ExpressJS, user specifies their paths without the '/' prefix
+    var endpoint = options.endpoint;
```
tejashah88: @bgjehu This should be `var endpoint = utils.normalizeApiPath(options.endpoint);`

bgjehu:
@dblock @tejashah88 
* clarified changes in CHANGELOG.md
* added normalization of param 'endpoint' for alexa.express() with util fn
* added upgrading guide for ExpressJS support in UPGRADING.md

not related to the commit, I testedthe new logic with alexa-app-server, swap the alexa-app node_module to my folk for all the examples (including the invalid ones), all test passed. However, not changing anything in alexa-app-server for its package.json is point to alexa-app@3.0.0 instead of 4.0.0. I would apply the associated changes to alexa-app-server once it starts pointing to alexa-app@4.0.0

dblock:
How do all our alexa-app-server tests pass when we're no longer attaching the router to the express instance? :)

bgjehu:
I had made related changes in alexa-app-server/index.js as well before I ran the test. They didn't just pass without changes for the usage of alexa.express() would be different.

index.js
```diff
@@ -560,25 +561,25 @@ alexa.app = function(name) {
...
-      throw new Error("You must specify an express instance to attach to.");
+    if (!options.expressApp && !options.router) {
+      throw new Error("You must specify an express app or an express router to attach to.");
     }
 
-    if (!options.router) {
-      throw new Error("You must specify an express router to attach.");
+    if (options.expressApp && options.router) {
+      console.warn("Since alexa-app@4.0.0, you are no longer required to pass both 'expressApp' and 'router'.");
+      console.warn("When both passed, 'alexa-app' would be attached to 'expressApp' instance.");
```
tejashah88: IMO, the warning message I would have done is something along the lines of "`'expressApp' and 'router' are both specified, using 'expressApp'...`". Generally, I want to give the user just the warning itself, and if they want to know more about it, they could look up [UPGRADING.md](https://github.com/alexa-js/alexa-app/blob/master/UPGRADING.md) for more information.

index.js
```diff
@@ -566,8 +566,7 @@ alexa.app = function(name) {
     }
 
     if (options.expressApp && options.router) {
-      console.warn("Since alexa-app@4.0.0, you are no longer required to pass both 'expressApp' and 'router'.");
-      console.warn("When both passed, 'alexa-app' would be attached to 'expressApp' instance.");
+      console.warn("Both 'expressApp' and 'router' are specified, attaching to 'expressApp' only.\nMore details on https://github.com/alexa-js/alexa-app/blob/master/UPGRADING.md");
```
bgjehu: @tejashah88 would this be better?

tejashah88: @bgjehu Sure. Let's see what @dblock has to say.

dblock: I think it should either be an exception or be backward compatible and produce a warning that this behavior is deprecated. 

If we ignore `router`, the user is not getting the behavior from the previous release and we just tell them "sorry, it's not going to work, but we are going to pretend like it does" :)

I actually would prefer a backward compatible version here with a warning, because that allows a lot more flexibility in combining versions of alexa-app-server and alexa-app.

What do you think?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Changelog
 
 ### 4.0.0 (Next)
+* [#176](https://github.com/alexa-js/alexa-app/pull/176): Made params `expressApp` and `router` optional for `alexa.express()` - [@bgjehu](https://github.com/bgjehu).
```
dblock: That's not what this change does, one of the two is required. So maybe "Express integration now requires either `expressApp` or `router`, but not both"?

README.md
```diff
@@ -89,14 +89,15 @@ app.express({ expressApp: express_app, router: express.Router() });
...
-* `endpoint` the path to attach the router to (e.g., passing `'mine'` attaches to `/mine`)
+* `expressApp` the express app instance to attach to
+* `router` the express router instance to attach to
+* `endpoint` the path to attach the express app or router to (e.g., passing `'mine'` attaches to `/mine`)
 * `checkCert` when true, applies Alexa certificate checking _(default: true)_
 * `debug` when true, sets up the route to handle GET requests _(default: false)_
 * `preRequest` function to execute before every POST
 * `postRequest` function to execute after every POST
 
+Either `expressApp` or `router` has to be specified so that alexa-app could be attached to.
```
dblock: has to be specified so that alexa-app could be attached to => `is required`. I think it's clear what that does already from above.

UPGRADING.md
```diff
@@ -39,6 +39,48 @@ app.intent("tellme", (request, response) => {
 });
 ```
 
+#### Changes to ExpressJS Support
+
+After changes, `alexa.express()` no longer requires both `expressApp` and `router` to be passes in. Instead, just pass in either an express app or router. And alexa-app would be bind to it.
```
dblock: That's not the whole change. Also it's unclear what `alexa` is, it should be `app`, and "to be passed in" is redundant. And the next sentence says the same thing. Maybe:

```
Express.js integration via `app.express()` now requires one of `expresApp` or `router`. When using a router, it will no longer be automatically mounted inside the express app and you may need to implement that outside of this call.
```

test/test_utils.js
```diff
@@ -0,0 +1,83 @@
+/*jshint expr: true*/
```
dblock: Lets delete unused functions here? The only one utilized is `normalizeApiPath`.

dblock:
I made some comments. Sorry to be a pest, I want to make sure we don't break a bunch of users.

index.js
bgjehu: @dblock make sense. Making it backward compatible tonight.

CHANGELOG.md
bgjehu: @dblock thanks for offering the words. I am not good with words and describing things! Will do the changes!

README.md
bgjehu: @dblock will do

UPGRADING.md
bgjehu: @dblock will do!

test/test_utils.js
bgjehu: @dblock yes sir

CHANGELOG.md
dblock: You're doing great, no stress!

bgjehu:
backward compatible tested with alexa-app-server
image

bgjehu:
@dblock @tejashah88 
Backward compatible tested with alexa-app-server
image

dblock:
I'm going to merge this, nice work on a spec, too!

I'd appreciate a follow-up pull request that:

* Deletes unused functions in utils.
* Slightly tweaks UPGRADING to say that the behavior with both is deprecated rather than removed.

dblock:
@tejashah88 Interested in cutting a minor release with this change?

bgjehu:
@dblock there is unused functional in utils? I saw the coverage for utils.js is 100%
will tweaks UPGRADING

dblock:
coverage means there're tests, but half the functions in utils aren't actually used, eg. `isValidDirectory`.

bgjehu:
right right right, silly me. I will double check and remove the unused

tejashah88:
@dblock We can do a release, however the [CHANGELOG](https://github.com/alexa-js/alexa-app/blob/master/CHANGELOG.md#312-february-16-2017) states that version `3.1.2` should be released today.

tejashah88:
We should either go to `3.1.1` or `4.0.0`, or wait until all relevant PRs are merged before deciding.

dblock:
The CHANGELOG is wrong, it shouldn't have been updated. The current release is still 3.1.0.

PR #180, opened 2017-02-14
Added utilities

tejashah88:
Needed for #177


dblock:
It should be part of #177 then, this is dead code as this PR stands.

PR #183, opened 2017-02-16
User-defined error handler should automatically resolve unresolved …

rickwargo:
…responses, update README to reflect behavior.

Add tests for error handler resolution and error messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,12 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
-### 3.1.2 (February 16, 2017)
+### 3.1.3 (February 16, 2017)
 
 * [#182](https://github.com/alexa-js/alexa-app/issues/182): Fix: user-defined error handler should automatically resolve unresolved responses, update README to reflect behavior - [@rickwargo](https://github.com/rickwargo).
 * [#183](https://github.com/alexa-js/alexa-app/pull/183): Add tests for error handler resolution and error messages - [@rickwargo](https://github.com/ajcrites).
 
-### 3.1.1 (February 16, 2017)
+### 3.1.2 (February 16, 2017)
 
 * [#184](https://github.com/alexa-js/alexa-app/pull/184): Implement slot defaults without exception, add test - [@rickwargo](https://github.com/rickwargo).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,14 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
-### 3.1.1 (February 16, 2017)
+### 3.1.2 (February 16, 2017)
 
 * [#182](https://github.com/alexa-js/alexa-app/issues/182): Fix: user-defined error handler should automatically resolve unresolved responses, update README to reflect behavior - [@rickwargo](https://github.com/rickwargo).
 * [#183](https://github.com/alexa-js/alexa-app/pull/183): Add tests for error handler resolution and error messages - [@rickwargo](https://github.com/ajcrites).
+
+### 3.1.1 (February 16, 2017)
+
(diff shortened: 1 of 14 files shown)


README.md
```diff
@@ -500,7 +503,8 @@ response.card({
 ## Error Handling
 
 Handler functions should not throw exceptions. Ideally, you should catch errors in your handlers using try/catch and respond with an appropriate output to the user. If exceptions do leak out of handlers, they will be thrown by default. Any exceptions can be handled by a generic error handler which you can define for your app. Error handlers cannot be asynchronous.
-
+The generic error handler, below, will automatically resolve the promise (request) with `response.send()` if it does not occur in the error handler.
+`response.fail(message, exception)` should only be called from a controlled calling environment, e.g. your own server, not AWS Lambda.
```
ajcrites: Why would this be the case? You should be able to use `response.fail` to force a failure response in any environment. Particularly this will trigger `.error` and `.post`.

index.js
```diff
@@ -357,6 +357,8 @@ alexa.app = function(name) {
       var handleError = function(e) {
         if (typeof self.error == "function") {
           self.error(e, request, response);
+          if (!response.resolved)
```
ajcrites: Include braces

index.js
```diff
@@ -366,11 +368,13 @@ alexa.app = function(name) {
           }
         }
         if (!response.resolved) {
-          if (e.message) {
-            response.fail("Unhandled exception: " + e.message + ".", e);
-          } else {
-            response.fail("Unhandled exception.", e);
-          }
+            if (e.message) {
```
ajcrites: Indent this block by two lines instead of four.

README.md
rickwargo: Maybe it needs a better description. Calling `response.fail()` in a Lambda environment produces an undesired message from Alexa.

index.js
rickwargo: I'm fighting with my editor and my default of 4 - missed this one.

README.md
dblock: I agree, maybe something explaining what happens on Lambda. Can be edited later.

test/test_alexa_app_unknown_type_request.js
```diff
@@ -21,25 +21,74 @@ describe("Alexa", function() {
...
+          };
+
+          var subject = testApp.request(mockRequest).then(function(response) {
+            return response.response.outputSpeech;
+          });
+          return expect(subject).to.eventually.be.rejectedWith("INVALID_REQUEST_TYPE");
         });
 
         it("fails with not a valid request", function() {
+          testApp.error = undefined; // may be defined in other tests, must unset
```
dblock: The test above sets it, wrap that into a context and have a `beforeEach` and `afterEach`, where you set the error before and unset it after. This way you don't need to rely on the order of tests.

test/test_alexa_app_unknown_type_request.js
```diff
@@ -21,25 +21,74 @@ describe("Alexa", function() {
 
         var mockRequest = mockHelper.load("unknown_type_request.json");
 
-        it("invokes a globally defined error function", function() {
+        it("invokes a globally defined error function that throws an error", function() {
           testApp.error = function(e, request, response) {
```
dblock: Similar to below, wrap error reset or create a new testApp instance every time.

rickwargo: Now create a new testApp instance and setting the error handler before each test.

test/test_alexa_app_unknown_type_request.js
rickwargo: Now create a new testApp instance and setting the error handler before each test.

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
+### 3.1.3 (February 16, 2017)
```
dblock: This should really happen when you do an actual release, but I'll let it slide ;)

CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
...
+
+* [#182](https://github.com/alexa-js/alexa-app/issues/182): Fix: user-defined error handler should automatically resolve unresolved responses, update README to reflect behavior - [@rickwargo](https://github.com/rickwargo).
```
dblock: No need to mention things like "update README to reflect behavior"

PR #184, opened 2017-02-16
Implement slot defaults without exception, add test

rickwargo:
No need to cause unwarranted exception about a missing intent when the intent is there, the default is not. This PR checks to see if the slot does not exist and returns the default value if not. Existing test covers this code path. Add test for getting default value when the request is actually missing the intent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
+### 3.1.2 (February 16, 2017)
+
+* [#184](https://github.com/alexa-js/alexa-app/pull/184): Implement slot defaults without exception, add test - [@rickwargo](https://github.com/rickwargo).
+
 ### 3.1.0 (February 13, 2017)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
+### 3.1.2 (February 16, 2017)
+
+* [#184](https://github.com/alexa-js/alexa-app/pull/184): Implement slot defaults without exception, add test - [@rickwargo](https://github.com/rickwargo).
```
dblock: Generally we don't need to note tests.

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 * [#174](https://github.com/alexa-js/alexa-app/pull/174): Always enfore strict header checking - [@tejashah88](https://github.com/tejashah88).
 * Your contribution here.
 
+### 3.1.2 (February 16, 2017)
```
tejashah88: Is there supposed to be a release today? Also, shouldn't it be `3.1.1`?

dblock: Yeah, that shouldn't be here. I'll cleanup.

PR #190, opened 2017-02-19
no junk drawer modules

mreinstein:
"util" modules tend to be a dumping ground for code. It's the digital equivalent of a junk drawer. Those things are valuable and stand on their own as proper named modules.

Also moved `to-ssml` and `normalize-api-path` to a `lib/` directory as-is common convention in many npm modules.


dblock:
thx

PR #197, opened 2017-03-01
Use callback parameter for default lambda function handler

### LINKED ISSUE #188: Switch to `callback` for lambda instead of the `context` object
Since node v4.3 on Lambda, using the [`callback`](http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html#nodejs-prog-model-handler-callback) parameter is preferred to indicate lambda function completion rather than using `context.succeed` or `context.fail`.

I would like to make this change. This would be a potentially breaking change that we can include in v4. The big difference is that the `context` methods don't wait for an empty event loop so any lambda functions relying on that functionality would be broken.

ajcrites:
Closes #188

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -83,6 +83,34 @@ express_app.use('/api', apiRouter);
 // now POST calls to /api/alexa in express will be handled by the app.request() function
 ```
 
+#### Lambda `callback` used by default over `context`
+Since version 4.0.0, the [`callback`](http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html#nodejs-prog-model-handler-callback)
+parameter that the lambda function takes is used to complete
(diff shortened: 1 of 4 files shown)


UPGRADING.md
```diff
@@ -83,6 +83,34 @@ express_app.use('/api', apiRouter);
 // now POST calls to /api/alexa in express will be handled by the app.request() function
 ```
 
+#### Lambda `callback` used by default over `context`
+Since version 4.0.0, the [`callback`](http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html#nodejs-prog-model-handler-callback)
```
dblock: Not a big deal, but since nobody reads these markdown files in markdown format there's no need to line-break them at 80.

PR #198, opened 2017-03-01
Added Heroku Button

tejashah88:
Any documentation suggestions are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 * [Examples](#examples)
     * [AWS Lambda](#aws-lambda)
     * [Express](#express)
+        * [Heroku Quickstart](#heroku-quickstart)
 * [API](#api)
     * [request](#request)
     * [response](#response)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,6 +139,13 @@ Either `expressApp` or `router` is required.
 A full express example is available [here](example/express.js).
 
 
+#### Heroku Quickstart
+
+Want to get started quickly with alexa-app and Heroku? Simply click the button below!
+
+[![Deploy to Heroku](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/tejashah88/alexa-app-example)
```
dblock: The template URL should be pointing to something in this repo, otherwise I don't think it belongs here.

dblock:
I'll close this since ^.

tejashah88:
@dblock Sorry for not responding for a while. Anyways, I'll make a new PR to merge my repo with the examples we have so far.

tejashah88:
So I've tried to integrate the button so that it will point to somewhere locally in alexa-app, (like /examples/heroku), but it seems as though Heroku doesn't like that.

image

What it's asking for is that the app.json must be on "the project **root directory**" ([see here](https://github.com/app-json/app.json#appjson-)), and if this file is not on the root directory, it simply won't work.

While I see what you are asking for, I think it should be designated to a separate repository, since Heroku requires a bunch of specific files and the main repository shouldn't be cluttered with these extra files, should there be any other similar integrations.

tejashah88:
So any chance of this PR being re-opened again?

dblock:
I think what you suggested made sense - ie. to create a separate demo app repo and add it in that repo's README.

tejashah88:
So I'll create a new PR, since I've deleted the original repository that contained the changes. If you want, I can transfer ownership of the [alexa-app-example](https://github.com/tejashah88/alexa-app-example) repository to the org and make the respective changes on the PR.

PR #202, opened 2017-03-11
Update first example for (AWS Lambda)

hidace:
Small update, but added in the require statement of the alexa-app module for slightly more clarity in the docs - (considering it's the first example). Thanks!


dblock:
Thanks.

PR #215, opened 2017-04-02
Fix Syntax of Session Details to match with Docs

danielstieber:
Hey, 

when I installed VoiceLabs analytics modul, I realized that it expects the original Request Body Syntax of a Session (in the Docs:https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#request-body-syntax).

Alexa-app is currently not providing the right syntax as you can see on the comparison of the files. 

I added an additional user tier, just as it is in the Alexa Docs to fix this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 3.0.1 (Next)
+
+* [#215](https://github.com/alexa-js/alexa-app/pull/215): Fix Syntax of Session Details to match with Docs - [@danielstieber](https://github.com/danielstieber).
+* Your contribution here.
+
 ### 4.0.0 (March 20, 2017)
 * [#134](https://github.com/alexa-js/alexa-app/issues/134): Asynchronous support purely through Promises, removed `return false`/callback support - [@ajcrites](https://github.com/ajcrites).
 * [#22](https://github.com/alexa-js/alexa-app/issues/22): Asynchronous support for `pre` and `post` - [@ajcrites](https://github.com/ajcrites).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,11 @@
 ## Changelog
 
-### 4.0.0 (Next)
+### 4.0.0 (March 20, 2017)
 * [#134](https://github.com/alexa-js/alexa-app/issues/134): Asynchronous support purely through Promises, removed `return false`/callback support - [@ajcrites](https://github.com/ajcrites).
 * [#22](https://github.com/alexa-js/alexa-app/issues/22): Asynchronous support for `pre` and `post` - [@ajcrites](https://github.com/ajcrites).
 * [#188](https://github.com/alexa-js/alexa-app/issues/188): Use `callback` to complete lambda functions rather than `context`. - [@ajcrites](https://github.com/ajcrites).
 * [#199](https://github.com/alexa-js/alexa-app/issues/199): Add support for `PlaybackController` events - [@tternes](http://github.com/tternes).
 * [#203](https://github.com/alexa-js/alexa-app/issues/203): Fix: utterances separated by space instead of tab - [@jmihalicza](http://github.com/jmihalicza).
-* Your contribution here.
 
 ### 3.2.0 (February 24, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## Changelog
 
-### 3.0.1 (Next)
+### 4.0.1 (Next)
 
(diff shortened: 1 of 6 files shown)


danielstieber:
I did the changelog, but I'm not sure how to test, since it the session syntax is already right in test fixtures (e.g. intent_request_airport_info.js).

This is my first PR, please tell me if I did something wrong/unusual :)

ajcrites:
This makes sense and I agree with the change, but I think that it is a potentially breaking change. For backwards compatibility, I think the `userId` and `accessToken` properties should be retained.

In a future version we can remove them as well as the deprecated session functionality.

index.js
```diff
@@ -262,8 +262,10 @@ alexa.session = function(session) {
     this.details = {
```
tejashah88: Technically speaking, since the hierarchy of the details object would match the session object, wouldn't it be easier to reduce the code to `this.details = session;`?

danielstieber: Well, that makes sense. For now I leave it like that because of the ongoing discussion.

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 4.0.1 (Next)
+
+* [#215](https://github.com/alexa-js/alexa-app/pull/215): Fix Syntax of Session Details to match with Docs - [@danielstieber](https://github.com/danielstieber).
```
dblock: Minor: lowercase everything except `Fix`, and make it `Fixed`. Also maybe be clearer about what session details we're talking about. I suggest: `Fixed session.details ...bla bla bla`.

index.js
```diff
@@ -262,8 +262,10 @@ alexa.session = function(session) {
...
       "new": session.new,
       "sessionId": session.sessionId,
-      "userId": session.user.userId,
-      "accessToken": session.user.accessToken || null,
+      "user": {
```
dblock: Shouldn't this just be `session.user` for future compatibility?

index.js
dblock: Yep, that would be the right thing to do, and is future-proof. I think since we're introducing a backward-incompatible change that makes sense.

dblock:
I think to merge this we need:

* A backward compatible `session.details.userId` and `accessToken` method that produces a warning that it's being deprecated.
* A paragraph in UPGRADING.

index.js
danielstieber: Hey, I'm stuck with the warning part, maybe you have a suggestion to solve this? I looked up a lot of stuff but still can't find what I need.

I solved the backwards compatibility like that (is that a proper way?):
```
  // Load whole session since 4.0.1 
    this.details = session;
    // Adding userdata again to be Backward compatible
    this.details.userId = session.user.userId; 
    this.details.accessToken = session.user.accessToken;
```

My idea to put in a warning was something like that:
```
  // Load whole session since 4.0.1 
    this.details = session;
    // Adding userdata again to be Backward compatible
    var olduserId = function() {
          console.error('WARNING: The syntax for userId has changed, please check alexa-app/UPGRADING.md for details.');
          return session.user.userId;
    };
    this.details.userId = olduserId();`
```
But now it gives out warning everytime someone uses the details object. It would be better to give a warning only when someone uses details.userId or details.accessToken.

Do you have an idea what I can do, or is there a name for that kind of warning, so I can search for a solution?

Thanks

tejashah88: The issue here is that the console will always log the error msg, simply because you are calling the function when constructing the request object. Unfortunately there's no easy way to generate a warning from using a specific property without refactoring major parts of the project or adding niche code to support this particular feature.

danielstieber: Ok, I was afraid that it would be complicated.. thank you for your answer @tejashah88 !

@dblock I will implement the backwards compatible way and a paragraph in UPGRADING for now. I will keep on looking for a way to implement the warning, but I'm afraid that might be beyond my current JS skills.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### 4.0.1 (Next)
 
+* [#215](https://github.com/alexa-js/alexa-app/pull/215): Fixed syntax of session.details to match with session as passed by amazon. - [@danielstieber](https://github.com/danielstieber).
```
dblock: Maybe "Fixed syntax of `session.details` to match with session object in the request"? Also remove the period inside (after amazon.).

dblock:
This looks good. Build needs to be fixed, please. I think we also need something in README about this whole jazz.

danielstieber:
Fixed the build. I had to edit the session syntax in a test file.

The README already says it passes the original session object. I just added the Amazon documentation link to the session object (saw it on the "response.card(Object card)" section of the README) so people can check the syntax, even when it changes in future Alexa updates. Hope that's fine.

dblock:
Merged via https://github.com/alexa-js/alexa-app/commit/498f6f97dc92e445565e5ddd98444d0bd3a63014. Thank you.

PR #218, opened 2017-04-04
Fix cert check test cases.

adrianba:
The error messages from alexa-verifier were changed in mreinstein/alexa-verifier@cde450b90078b52eb45a5e865f3083d30fa382da causing integration tests to fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.0.1 (Next)
 
 * [#219](https://github.com/alexa-js/alexa-app/pull/219): Preserve session when clearing non-existent attribute - [@adrianba](https://github.com/adrianba).
+* [#218](https://github.com/alexa-js/alexa-app/pull/218): Fix cert check test cases - [@adrianba](https://github.com/adrianba). 
 
 ### 4.0.0 (March 20, 2017)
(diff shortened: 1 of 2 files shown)


tejashah88:
Good catch! When it's all fixed, I'll merge soon.

adrianba:
Travis-CI isn't picking up the new version, possibly because the `.travis.yml` file caches the node_modules folder.

dblock:
I deleted caches and restarted the build, which passed mostly, but not for 0.12. Can you please look at that one?

There's probably a way to lock the dependency for dev purposes only via `devDependencies`?

adrianba:
I've rebased the PR to resolve the CHANGELOG conflict.

[`alexa-verifier`](https://github.com/mreinstein/alexa-verifier/) does not appear to support Node 0.12 any longer (its tests fail when run with 0.12 and the [`package.json`](https://github.com/mreinstein/alexa-verifier/blob/master/package.json#L27) requires node >= 1).

tejashah88:
If alexa-verifier doesn't support node v0.12, then this library shouldn't support it as well. I'll make a PR.

dblock:
It looks like tests are passing with 0.12 tho.

adrianba:
No, the tests [failed on the check-in](https://travis-ci.org/alexa-js/alexa-app/builds/218491325) with Node 0.12.

dblock:
Ah I see, I missed a red thing somehow. Mind taking it out of .travis.yml please in a PR?

adrianba:
See #220.

PR #219, opened 2017-04-04
Preserve session when clearing non-existent attribute.

adrianba:
session.clear("xyz") will delete all session attributes if "xyz" does not exist.


dblock:
Oh good catch!

PR #220, opened 2017-04-04
Remove node 0.12 from Travis-CI.

adrianba:
Dependencies for alexa-app no longer work on Node 0.12 so remove node 0.12 from test matrix. Run extra steps with Node 1 instead. Also updated engine version in `package.json`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 * [#219](https://github.com/alexa-js/alexa-app/pull/219): Preserve session when clearing non-existent attribute - [@adrianba](https://github.com/adrianba).
 * [#218](https://github.com/alexa-js/alexa-app/pull/218): Fix cert check test cases - [@adrianba](https://github.com/adrianba). 
-* [#220](https://github.com/alexa-js/alexa-app/pull/220): Remove Node 0.12 from Travis-CI - [@adrianba](https://github.com/adrianba). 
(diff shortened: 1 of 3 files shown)


tejashah88:
@dblock @adrianba Since we are removing testing support for node v0.12, would it make sense to just drop node v0.12 support for this library entirely?

dblock:
Yes, please, do remove node 0.12 support if we can't run tests anymore. Also we do want a green build whatever that takes ;)

adrianba:
Working on it - seeing some odd errors so will set-up Travis against my fork and figure out what is going on.

adrianba:
Okay, seems to be working now - also updated min version in `package.json`.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * [#219](https://github.com/alexa-js/alexa-app/pull/219): Preserve session when clearing non-existent attribute - [@adrianba](https://github.com/adrianba).
 * [#218](https://github.com/alexa-js/alexa-app/pull/218): Fix cert check test cases - [@adrianba](https://github.com/adrianba). 
+* [#220](https://github.com/alexa-js/alexa-app/pull/220): Remove Node 0.12 from Travis-CI - [@adrianba](https://github.com/adrianba).
```
dblock: I think this should say "No longer compatible with Node 0.12".

adrianba: Done.

dblock:
Thanks @adrianba.

PR #226, opened 2017-04-08
Add properties for device ID and consent token

martincostello:
Add properties to the Alexa request to easily access the device ID and content token, for example for use after returning a permission card to access the device's address.

  * [Getting the device ID and consent token](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/device-address-api#get-the-consent-token-and-device-id)
  * [Getting a device's address](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/device-address-api#getCountryAndPostalCode)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ Temporary Items
 
 # IDE
 .idea
+.vs
 .vscode
 
 # NPM
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 4.0.1 (Next)
 
-* [#226](https://github.com/alexa-js/alexa-app/pull/226): Added properties for device ID and consent token - [@martincostello](https://github.com/martincostello).
+* [#226](https://github.com/alexa-js/alexa-app/pull/226): Added properties for API endpoint, consent token and device ID - [@martincostello](https://github.com/martincostello).
 * [#223](https://github.com/alexa-js/alexa-app/issues/223): Added support for `AskForPermissionsConsent` cards - [@ericblade](https://github.com/ericblade).
 * [#219](https://github.com/alexa-js/alexa-app/pull/219): Preserve session when clearing non-existent attribute - [@adrianba](https://github.com/adrianba).
 * [#218](https://github.com/alexa-js/alexa-app/pull/218): Fix cert check test cases - [@adrianba](https://github.com/adrianba). 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.0.1 (Next)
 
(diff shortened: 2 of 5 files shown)


martincostello:
I'll also look at doing a PR to alexa-app-server to put these properties in the sample requests for testing.

martincostello:
Have also added support for ```apiEndpoint``` ([docs](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#system-object)) which is needed in conjunction with ```deviceId``` and ```consentToken``` for the [Device Address API](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/device-address-api#base-uris-and-geographic-location-of-the-skill).

dblock:
I think this looks fine, but before I merge this ... I am worried that we're plucking all kinds of properties out of the context perpetuating the bad pattern of "flattening" `context` into properties. So the question is why do we want these at the top level and why we aren't accessing them directly as `this.data.context.System.device.deviceId`?

If these methods were to provide more useful logic I would want to wrap `context` into an object with additional methods, but they don't seem to. Also curious what others like @ajcrites or @tejashah88 think about this.

martincostello:
That's fair, I did notice the comment in a similar PR about moving stuff around after I opened this. I assumed that having some "helper" properties for these might be useful as it simplifies the integrator experience rather than having to drop-down to the "raw" request to pull out the two values from the fairly deeply-nested hierarchy.

dblock:
I'm just worried that we're going to get a bunch of unrelated properties at the top level.

What about wrapping `context` and providing actual objects for some of these that are `null`-proof as well?

martincostello:
So something like adding a new ```getContext()``` function (like ```getSession()```), which then returns an object that is a fully initialised representing of the [Context](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#context-object) object, but with the values copied over from the raw request?

dblock:
I think it doesn't need to copy anything, it would be a "view" on the data, but it would consistently return `null` when nested properties don't exist.

martincostello:
Coming back to this, so I guess to do something like [this](https://github.com/alexa-js/alexa-app/blob/498f6f97dc92e445565e5ddd98444d0bd3a63014/index.js#L218):

```js
var context = new alexa.context(json.context);
this.getContext = function () {
  return context;
};
```

Where `alexa.context` is similar to the implementation of `alexa.session`, except entirely read-only?

dblock:
Right @martincostello

martincostello:
Having looked at changing this PR to the above, given the context is already available and as it's conceptually read-only (unlike session) I don't think there's much value in adding some new methods to wrap it all up.

If they're not convenience properties at the top like the others, which you're trying to move away from, I think you'd end up referencing as deep as you would directly into `context` as it is now, so I'm just going to close this.

PR #227, opened 2017-04-11
Added heroku button section

tejashah88:
From #198.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,7 +143,7 @@ A full express example is available [here](example/express.js).
 
 Want to get started quickly with alexa-app and Heroku? Simply click the button below!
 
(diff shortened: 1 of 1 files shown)


dblock:
I am OK merging this, but I am worried that the code in that example will not be the latest code here. It's also deploying alexa-app-server, so feels like two levels of indirection. Why isn't the heroku deploy button *in that project*?

Transferring the example repo probably makes sense. I believe you have enough permissions for that otherwise lmk.

tejashah88:
@dblock I'll update the repo to the latest version of alexa-app. Where exactly is alexa-app-server used, because from the corresponding [package.json](https://github.com/tejashah88/alexa-app-example/blob/master/package.json#L25), the only dependencies are alexa-app, ejs, and express.

tejashah88:
I've also transferred the repo to the org and updated the links for this PR and that repo.

PR #236, opened 2017-05-05
Update ssml.cleanse to stop removing spaces before decimals

aminimalanimal:
alexa-app was unexpectedly stripping spaces that occur before periods. Since periods are also decimals, this produced issues with the expected output of our bartender app's card text.

Things like:
> Ingredients: 2 oz Tanqueray No. TEN Gin, .5 oz Heavy cream, .5 oz Fresh lemon juice, .5 oz Fresh lime juice, .75 oz Simple syrup (one part sugar, one part water), 3 dashes Orange flower water, 1 Fresh egg white, and Club soda

were showing up as:
> Ingredients: 2 oz Tanqueray No. TEN Gin,.5 oz Heavy cream,.5 oz Fresh lemon juice,.5 oz Fresh lime juice,.75 oz Simple syrup (one part sugar, one part water), 3 dashes Orange flower water, 1 Fresh egg white, and Club soda

I think that this sort of cleansing should probably not be done at all, instead letting app developers to have control over their output, but this PR only removes the item that affected us. Perhaps the others will always be ideal to strip spaces for, but it's a hard thing to predict—after all, our use-case wasn't immediately obvious. :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.0.1 (Next)
 
+* [#236](https://github.com/alexa-js/alexa-app/pull/236): Update `ssml.cleanse` to stop removing spaces before decimals - [@aminimalanimal](https://github.com/aminimalanimal).
 * [#215](https://github.com/alexa-js/alexa-app/pull/215): Fixed `session.details` to match with session in the request object - [@danielstieber](https://github.com/danielstieber).
 * [#223](https://github.com/alexa-js/alexa-app/issues/223): Added support for `AskForPermissionsConsent` cards - [@ericblade](https://github.com/ericblade).
 * [#219](https://github.com/alexa-js/alexa-app/pull/219): Preserve session when clearing non-existent attribute - [@adrianba](https://github.com/adrianba).
--- a/lib/to-ssml.js
+++ b/lib/to-ssml.js
@@ -31,7 +31,7 @@ var ssml = {
   },
   cleanse: function(str) {
     // <p> is left in place to support intended HTML output
-    return str.replace(/<\/?(speak|break|phoneme|audio|say-as|s\b|w\b)[^>]*>/gi, " ")
+    return str.replace(/<\/?(speak|break|phoneme|audio|say-as|s\b|w\b)[^>]*>/gi, "")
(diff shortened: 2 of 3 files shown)


aminimalanimal:
Apologies—I'm uncertain why the Travis CI build is failing.

dblock:
The build failure is clearly about the changed behavior, so please run and fix the tests and maybe add more? LMK if you need help.

aminimalanimal:
I just uploaded a change that passes the existing tests. Uncertain what the intention behind replacing `speak`/etc with a space was in the first place, so I don't know if this could potentially negatively impact something else.

aminimalanimal:
btw, when I clicked on the travis-ci build bot earlier, it didn't provide me with any sort of _failed test_ output like running `npm run test` does; it just had some generic failure messages about environment variables or something.

ericblade:
umm.. a possible? point? might be that SSML is intended for the voice output, not the card output.  Or is this some sort of auto-generated card, not a requested card?

In any case, I think that Alexa takes hints from the output punctuation, so not sure why we're pulling things out of it ..

dblock:
The way the change is implemented right now needs additional tests that would fail with the old code, please. Otherwise it's easy to reintroduce a regression.

When I code reviewed this first the output changes in tests made sense, now they are back to the original version, which is probably not what was intended?

aminimalanimal:
@ericblade The `cleanse` method is run to “remove all SSML to keep the card clean,” so even though this is in a file named `to-ssml`, it's really only meant to be used when converting from SSML.

aminimalanimal:
@dblock Looking into...

aminimalanimal:
> When I code reviewed this first the output changes in tests made sense, now they are back to the original version, which is probably not what was intended?

It seems to me like the expectations behind all existing tests are still valid, so I'm not sure what you mean. I'm still seeing the expected output that I made the change for (which broke the tests), and removing the space that was added while stripping the SSML tags un-broke the tests.

I've added a test to prevent regression.

Thanks for your patience and explanations in this matter. I haven't done much TDD, so the thought to add a test to ensure that my change stays in place didn't occur to me.

dblock:
Looks good, thanks. Merging.

PR #247, opened 2017-06-08
Update README.md to contain AMAZON Intent Example

acucciniello:
- I was not sure how to do this to start, until I tried this,
      Someone with less experience developing Skills might not have
      figured it out right away.
    - Meant for Intents  with form of  AMAZON.IntentName
    - It is necessary for people to have some of these intents if they
      are trying to publish their skills.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Changelog
 
 ### 4.0.2 (Next)
-* Your contribution here
+* [#247](https://github.com/alexa-js/alexa-app/pull/247): Update README.md to contain AMAZON Intent Example - [@acucciniello](https://github.com/acucciniello).
 
 ### 4.0.1 (May 17, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## Changelog
 
 ### 4.0.2 (Next)
-* [#247](https://github.com/alexa-js/alexa-app/pull/247): Update README.md to contain AMAZON Intent Example - [@acucciniello](https://github.com/acucciniello).
+* Your contribution here
 
 ### 4.0.1 (May 17, 2017)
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## Changelog
 
 ### 4.0.2 (Next)
-* Your contribution here
```
dblock: You can put this back, and for this we don't need an entry really.

README.md
```diff
@@ -293,6 +293,23 @@ app.intent("vacation", function(request, response) {
...
+
+Amazon has specific intents that have to do with basic functionality of your skill that you must add.  Some examples of this are `AMAZON.HelpIntent`, `AMAZON.StopIntent`, and `AMAZON.CancelIntent`.  Here is an example of how you would specify that type of intent.
+
+```javascript
+app.intent('AMAZON.StopIntent',{
+  'slots': {},
+  'utterances': []
+}, function (request, response) {
+  	var stopOutput = 'Stopping your Request and Exiting Skill'
+  	response.say(stopOutput).send()
```
dblock: This is a good opportunity to be funny. Just saying :) Like `response.say("I'll be back.")`.

README.md
```diff
@@ -293,6 +293,23 @@ app.intent("vacation", function(request, response) {
...
+  	return
+})
+```
+
+You do not need to pass any utterances or slots into these intents.  Also when specifying the name of the intent just use the exact name Amazon gives you.
```
dblock: gives you -> provides

README.md
acucciniello: Great idea! I stole your idea and added my own spin

ericblade: Probably should follow the style of the previous example section too, dblock uses double quotes ..

README.md
ericblade: May want to provide a link to the page on Amazon about providing utterances for built-in intents, i believe that that is an option ..

dblock:
Merged, thanks.

PR #250, opened 2017-06-22
Add support for the Dialogue directive

sbonami:
So that intents can have additional context around schema (slots and utterances) as well as new functionality, an `alexa.intent` object was added. Additionally, an `alexa.dialogue` object was created to contain introspection logic. To benefit from the addition of the logic and ease development, auto-delegation was also added.


sbonami:
@ericblade I believe you mentioned an interest in looking at this chunk of code. I'd welcome feedback!

index.js
```diff
@@ -270,6 +270,13 @@ alexa.request = function(json) {
   };
 };
 
+alexa.intent = function(name, schema, handler) {
+  this.name = name;
+  this.handler = handler;
+  this.slots = (schema && typeof schema["slots"] != "undefined") ? schema["slots"] : null;
```
ericblade: !== ?

index.js
```diff
@@ -460,8 +461,8 @@ alexa.app = function(name) {
       if (!response.resolved) {
         if ("IntentRequest" === requestType) {
           var intent = request_json.request.intent.name;
-          if (typeof self.intents[intent] != "undefined" && typeof self.intents[intent]["function"] == "function") {
-            return Promise.resolve(self.intents[intent]["function"](request, response));
+          if (typeof self.intents[intent] != "undefined" && typeof self.intents[intent].handler == "function") {
```
ericblade: === ?

ericblade:
OK, I think I need to brush up on the Amazon docs for both Dialog and Directives :-)

So, I don't see any issues in the code, other than a couple of spots where I'd use === or !== without knowing the specifics of the Dialog and Directives APIs.

In my opinion, it'd be good to switch to the American English form of "Dialog", as that is what is used in the Amazon docs as well.

I'm a little confused, though, in the first commit how "        for (key in intent.schema.slots) {		 " becomes "        for (key in intent.slots) {" .. and still works, and we still have intent.schema.  Not sure if it's lack of sleep, or if I'm blindly missing something.  I assume you've tested the basic functionality.  Have you also checked the ?schema and ?utterances functions?

Otherwise, I'd just like to see a link in the docs pointing to the pages for Dialog and Directive, and perhaps an example of how to use them (i think the docs have extensive examples of the basics)

... and then i need to re-write all my code, because I basically wrote this entire Dialog system in my own skill code, about 3 weeks before Amazon came out with it :-D

sbonami:
@ericblade Great feedback. I'll tweak the english, README, and equality check. As for the `alexa.intent#slot` question, the `intent` is now an object with a slot attribute. No more `schema` attribute on an `intent`. It also seems as though the schema and utterance functions are [fully tested](https://github.com/alexa-js/alexa-app/blob/master/test/test_alexa_app_utterances_schema.js), but I can manually verify as well.

dblock:
This is great. Definitely lets make this US-EN because America first :)

dblock:
Missing file, check out the build pls.

sbonami:
@dblock Missed those in the rename. Fixed and ready to go!

dblock:
Merged, thanks.

sbonami:
@dblock Actually, should this be under `4.1`?

dblock:
Sure, lets increment.

PR #256, opened 2017-06-30
Added specific examples of required intents to README

whatsPaulUpTo:
This should help give a little more clarity on how to implement these necessary intents into an Alexa skill, specifically how the .HelpIntent differs in requiring the session to be left open.


dblock:
Great, thanks!

PR #258, opened 2017-07-05
Resolve: Updating request.type() in app.pre is ignored in later stages

samueljr:
There is a small bug in the current repository where if you change the value of what request.type() would return in your app.pre() function, it is not picked up before the request is acted on.  By relearning what requestType is later, this bug can be fixed.

### DIFF
--- a/index.js
+++ b/index.js
@@ -491,6 +491,7 @@ alexa.app = function(name) {
       return prePromise;
     }).then(function () {
       if (!response.resolved) {
+        requestType = request.type();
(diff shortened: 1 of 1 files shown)


samueljr:
Probably no need to update CHANGELOG.mg since this is so minor.

dblock:
Actually it's a bug, so it needs a test and a CHANGELOG, for sure. Thank you.

samueljr:
#259

PR #264, opened 2017-07-26
Adding support for the Display.ElementSelected type request. This wil…

jstreb:
…l look for function call "displayElementSelected" and call if it is defined. If it is not defined it will throw an error of "NO_DISPLAY_ELEMENT_SELECTED_FUNCTION".

Additionally this populates the token that is associated with this request on the top level request object on a property called 'selectedElementToken'.

### DIFF
--- a/index.js
+++ b/index.js
@@ -233,6 +233,11 @@ alexa.request = function(json) {
   if (this.data.request && this.data.request.intent) {
     this.confirmationStatus = this.data.request.intent.confirmationStatus;
   }
+
+  if (this.data.request && this.data.request.type === 'Display.ElementSelected' && this.data.request.token) {
+    this.selectedElementToken = this.data.request.token;
+  }
+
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -525,6 +536,13 @@ alexa.app = function(name) {
           if (typeof playbackEventHandlerObject != "undefined" && typeof playbackEventHandlerObject["function"] == "function") {
             return Promise.resolve(playbackEventHandlerObject["function"](request, response));
           }
+        } else if ("Display.ElementSelected" === requestType) {
+          if (typeof self.displayElementSelectedFunc === "function") {
+            return Promise.resolve(self.displayElementSelectedFunc(request, response));
+          } else {
+            console.log('Throwing NO_DISPLAY')
```
brandonaaskov: Do you want this here still?

jstreb: no I do not.

PR #265, opened 2017-07-27
Adding support for the Display.ElementSelected type request. This wil…

jstreb:
…l look for function call "displayElementSelected" and call if it is defined. If it is not defined it will throw an error of "NO_DISPLAY_ELEMENT_SELECTED_FUNCTION".

Additionally this populates the token that is associated with this request on the top level request object on a property called 'selectedElementToken'.


jstreb:
Thanks for the quick review and merge!

Issue #148, opened 2017-02-03
Typescript Implementation and Typings


RobertTheProfessional:
Any need for a Typescript implementation of this library? I rewrote the library in Typescript for a work project and removed the try/catches and implemented Promise rather than returning functions. There have been a few updates to the library since I've rewritten it but I can get source up to speed pretty quickly.

dblock:
I see people do typescript everywhere around me. I guess we should look at how much better that is vs. what we have. Curious what other people active in this project like @mreinstein and @tejashah88 think.

tejashah88:
While I don't normally use TypeScript on a daily basis, it would be nice to have such a library. The only issue that I foresee is maintaining similar changes between this library and the TS one.

dblock:
I think @RobertTheProfessional is proposing replacing alexa-app by a TypeScript library and compile that down to JS that runs on node. I am by no means saying that's what we should do, but I find arguments in https://github.com/artsy/emission/issues/18 or https://github.com/orta/systems-theory compelling.

mreinstein:
totally opinion based, but I am absolutely not interested in typescript.

RobertTheProfessional:
@tejashah88 My TypeScript implementation practically mirrors the functionality of the current JavaScript implementation with a few improvements on top; but the exclusion of the AudioPlayer functionality. I can roll back a few of my specific enhancements and finish adding the AudioPlayer code to ensure that it compiles down to a near perfect match.

@dblock @mreinstein At the very least I can generate the Typings from the TypeScript implementation for the purpose of autocomplete. Personally, TypeScript with alexa-app has been a great help in the development of my larger Node-RED plugin and quick proof of concepts, since it can be converted between ES5, ES6 and ES7 by configuring the compiler options.

mreinstein:
I'm certainly not going to tell you how to run your own projects; feel free to use whatever you find useful which makes you productive.

This project will never adopt typescript because:

a) the best written javascript modules are written in javascript. Transpiled source that produces js is never as high a quality as hand written, carefully tuned js written by people.

b) I believe the premise of typescript is flawed (that the only way to write large scale projects is with strong typing)

c) I don't want to have to know both the transpiler language and the target language.

d) javascript has a much higher pool of talent to pull from.

e) I don't want an additional compile step to complicate the build process unless it provides a lot of value.

dblock:
[...]

PR #266, opened 2017-07-27
TypeScript type definitions

lazerwalker:
Hey! I've got a draft of a TypeScript definition file for y'all.

If I'm understanding TypeScript correctly, this file can either live in this repo (preferred) or as part of DefinitelyTyped. Understand totally if you don't want this in here, but poking through the issue history suggests you're amenable to this approach :)

Afraid I'm still relatively new to TS, so apologies if there's stuff that's either not idiomatic or flat-out doesn't work (my testing strategy so far has been "I can compile my TS app that consumes this library without the compiler yelling at me". It currently passes that relatively low bar).

Before I get this in, I'd like to figure out what the testing situation looks like (since it seems like everything in the DefinitelyTyped library has some sort of tests attached to them), but wanted to open this for discussion sooner rather than later in case there's e.g. something obvious I messed up in how this actually represents the library.

It's also missing a bunch of more finicky stuff (e.g. all the raw JSON that lives within `response.data`). Working on filling it out, but, again, figured I'd start the conversation earlier.

Cheers!


dblock:
We've discussed #148. Lets re-read that first?

lazerwalker:
Yep, read that before I started this and just re-read now.

If I'm interpreting correctly, that was largely a discussion about whether e.g. rewriting the whole thing in TS was appropriate (including the current existence of a full-on TS fork as an exercise).

Seems to me that adding in a typing file is a bit tangential to any of that discussion, yeah? There's a potential conversation to have surrounding the additional maintenance cost of updating the type definition whenever the API changes, and maaaaybe some politics surrounding whether that type definition lives in this repo or on DefinitelyTyped, but those are very different beasts than talking about the sort of large-scale refactors talked about in that issue.

dblock:
Thanks @lazerwalker!

The first question is where does this live. I think it's really about how others do it and I am definitely not an expert. Curious what @orta thinks? Any other libraries out there doing one or the other?

A middle ground is that there's a new repo that depends on alexa-js and provides typescript definitions. It would then be able to depend on a specific version of alexa-js and we would have explicit api upgrades version-after-version. Of course it's more maintenance, but it can be done by people more motivated by TS.

The last question is maintenance and tests. We definitely want tests if code goes here or elsewhere and we want to run them. At the very least there needs to be a test that compares the API and the TS signatures and makes sure they didn't diverge.

orta:
Hello all, whether it fits in DTyped or in here really depends on how mature you think your public API is today. 

For something that is continually evolved DT probably is the better option, as updates to the types can ship outside of updates to the library. 

A good way to test them is to "create some example apps" in the repo and use [dtslint](https://github.com/Microsoft/dtslint) on them. These apps would get updated when your public API changes and the dtslint would fail if the types don't match up to what the example apps look like.

dblock:
@lazerwalker LMK what you think next steps are here. 

Maybe @ajcrites, @tejashah88, @matt-kruse  and @mreinstein want to comment too.

lazerwalker:
On my end, there are a few technical things:
- [ ] Flesh out type definitions for some of the helper objects that are currently missing (e.g. the raw `request.data` blob)
- [ ] Add tests

I'd say the latter is a prereq to getting this merged in anywhere, whereas the former wouldn't be the end of the world if it got done later. What sorts of tests exist might also depend on whether these live in-repo or on DT, will investigate when I sit down to do that. I'm a bit slammed with other work this week, but I'm shooting to bang these out later this week or over the weekend.

Other than that, the question still remains of whether it lives in DT vs here. As an outsider/newcomer, this seems like a relatively stable API, which suggests to me that living within this repo is the right place. This feels like something that needs input from contributors with more historical knowledge, though. Thoughts, @ajcrites @tejashah88 @matt-kruse @mreinstein et al?

If it should live within the repo: once my tests/etc are done, I think the main bit of housekeeping required is making sure any tests I write are wired up to CI builds (presumably wired up as other code tests are instead of e.g. as a Danger task, but that may depend again on what form the tests actually take).

If it should live in DefinitelyTyped, this PR can be closed out and the only real change within the repo would be maybe a note in the README.

ajcrites:
I think living in-repo is generally better than DefinitelyTyped -- the latter was created to allow third parties to create type definitions for repos that couldn't/wouldn't support them in-repo.

Other than that I don't have any specific suggestions.

lazerwalker:
Just pushed a whole bunch of stuff!

* The type definitions now cover pretty much the entire API, including all of the objects being passed straight through from Amazon JSON. 
* There are tests via `dtslint`, which run successfully on Travis. One thing to note is that a bunch of tests on Travis which were formerly running on node v1 are now running on the latest version of node. We're still running the vanilla test suite on node v1.

To say the current tests are bare-bones is an understatement, but it's a good proof-of-concept for "we're actually running tests that type-check".

dblock:
I'd be down with merging this. Want to write up something in the README about how to use it, and CHANGELOG?

lazerwalker:
Already added to the CHANGELOG. Added a few lines to the README and CONTRIBUTING.md.

Didn't actually say anything about how to use it as a consumer, but it shouldn't need any instructions. E.g. anyone who opens a project in VS Code that has this as a dependency will get working autocomplete without them having to do anything.

types/index.d.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  express: (options: ExpressOptions) => void;
+
+  intent: (intentName: string, schema: IntentSchema, handler: RequestHandler) => void;
+}
+
+export class request {
+  type: () => "LaunchRequest"|"IntentRequest"|"SessionEndedRequest";
+
+  // return the value passed in for a given slot name
```
orta: Is this enough for showing in auto-completion? Might want to be `/** thing */`

lazerwalker: Ah, fascinating. When I first poked at doing `/** etc */`, I couldn't get those comments to show up in VS Code for me, so I figured leaving the comments as-is didn't matter (the whole structure is just taken from the API docs in the README). 

Seems like it was a PEBKAC, though — `*doc`-style comments are totally showing up for me now. I'll go through and tidy that up.

lazerwalker:
Converted all comments to JSDoc-style ones for VS Code, added a bunch more comments (nothing myself really, mostly just things that were easily copy/pastable from index.js or the README), and also fixed one or two other small issues that sweep revealed.

.travis.yml
```diff
@@ -4,15 +4,17 @@ sudo: false
 
 matrix:
   include:
-    - node_js: 1
+    - node_js: node
       script:
         - npm run-script danger
         - npm run-script lint
         - npm run-script test-with-coverage
+        - if [[ $TRAVIS_NODE_VERSION == "8" ]]; then npm install dtslint; npm run dtslint; fi
```
dblock: This should be a separate `matrix: include: ` all by itself instead of an `if`.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [#248](https://github.com/alexa-js/alexa-app/pull/248): Add support for intent and slot `confirmationStatus` - [@sbonami](https://github.com/sbonami).
 * [#250](https://github.com/alexa-js/alexa-app/pull/250): Add support for handling and/or delegating Dialog - [@sbonami](https://github.com/sbonami).
 * [#258](https://github.com/alexa-js/alexa-app/pull/258): Fix ability to edit request.type() in app.pre - [@samueljr](https://github.com/samueljr).
+* [#266](https://github.com/alexa-js/alexa-app/pull/266): Add TypeScript definition file - [@lazerwalker](https://github.com/lazerwalker)
```
dblock: I'd remove "file" and make it "definitions"

dblock:
Should the file be called `amazon.d`, the platform is called Alexa, but maybe I am missing something.

types/test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+app.intent('intentName', {
+  slots: {},
+  utterances: ['do a thing']
+}, (request, response) => {
+  response.say("Say a thing!")
+    .card("Show a card", "Wow, you showed a card!")
+    .card({type: "Simple"})
+    .shouldEndSession(false, "Reprompt!")
+});
```
dblock: To be consistent I think this should be in `tests`, not in `types`.

dblock:
Okay, so one important thing missing is a minimum doc on how to use this. I'm a noob, I have the library that I got in my project from npm, and I use VSCode. Now what?

.travis.yml
lazerwalker: Realizing that, since this script is just running on the latest node (which'll be >= 8), we can probably just remove the if statement entirely and directly `npm install`. Seem reasonable?

types/test.ts
lazerwalker: Seems like including the tests within the `types` folder is a `dtslint` convention. Maybe an @orta question?

lazerwalker:
If you open a project in VSCode that includes the library via npm, there's literally nothing you need to do.

If you type e.g. 

```js
const alexa = require('alexa-app');
const app = alexa.app(
```

You should get an IntelliSense popup telling you about the arguments of the `app` function. Nothing else required. 

Not sure that actually warrants any documentation?

.travis.yml
dblock: Yes.

dblock:
If it's magical, I'm down with no docs. LMK when you're done with the other nitpicks.

lazerwalker:
@dblock Should be good!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 * [#248](https://github.com/alexa-js/alexa-app/pull/248): Add support for intent and slot `confirmationStatus` - [@sbonami](https://github.com/sbonami).
 * [#250](https://github.com/alexa-js/alexa-app/pull/250): Add support for handling and/or delegating Dialog - [@sbonami](https://github.com/sbonami).
 * [#258](https://github.com/alexa-js/alexa-app/pull/258): Fix ability to edit request.type() in app.pre - [@samueljr](https://github.com/samueljr).
+* [#266](https://github.com/alexa-js/alexa-app/pull/266): Add TypeScript definitions - [@lazerwalker](https://github.com/lazerwalker)
```
dblock: This is missing a period in the end ;)

dblock:
Merged via https://github.com/alexa-js/alexa-app/commit/e3b966d7191639eda27464a0c379dd21cb395738. Thank you!

dblock:
@lazerwalker Want to join the maintainers crew and make a release with this?

jontg:
@dblock @lazerwalker you guys messed the signature for `lambda()` up, which threw me off this afternoon.  Care to review https://github.com/alexa-js/alexa-app/pull/269 ?

henry74:
Has this been released?  I'm using alexa-app with VSC and I'm not getting the typescript autocomplete kicking in.  I'm on 4.2.0.  Autocomplete works with my other projects.  I'm using it in index.ts file.

dblock:
I believe that yes, @henry74, but see https://github.com/alexa-js/alexa-app/blob/master/CHANGELOG.md in case something else has been since fixed.

Issue #255, opened 2017-06-29
Intent confirmation


adriendomoison:
Hello,

I was wondering how was the intent confirmation integrated with the framework? I couldn't find anything about it, maybe it is still under development?
Here is the feature I am referring to in the Alexa Skill Builder (beta).

https://goo.gl/photos/4jdyjChyLEiJ4jbN7

Thanks !

Ps: I love this framework!!

dblock:
Doesn't look like we support anything explicitly for this, please contribute.

ericblade:
I thought we added the confirmation stuff with the conversation stuff right
after the directives

dblock:
@ericblade Did we? Link?

dblock:
You mean https://github.com/alexa-js/alexa-app/pull/250? Does that do what you're describing @adriendomoison ?

ericblade:
I guess "we" (i mean the people that aren't me that are contributing with more than conversation right now :) ) did the confirmation before the dialog.  https://github.com/alexa-js/alexa-app/commit/b4e110647d167f06d56269186298d2d66bdd7f8c  #248

adriendomoison:
I am not sure what is the dialog feature, the Alexa skill builder feature I am talking about allows you to preset some sentences that alexa can say, so I guess the framework would have to be able to retrieve this list? I am not familiar with this confirmation feature yet so I am not sure about how it is suppose to work exactly.

kevingilbert100:
I've made my own custom router, I won't provide the exact code around it. However, I suggest you do the same thing. Amazon suggests using the session object to help with dialog like interactions. Here is an example of how I have my router structured.

```
router
  .intent({
    name: 'someIntent',
    utterances: ['do yo thing'],
    requiresPin: true,
    controller: someController,
    confirm: someConfirmController,
    reject: someRejectController
  });
```

So basically you have the AMAZON.YesIntent and AMAZON.NoIntent wrapped to check for a session key / value to know which controller to hit.

You can engineer the rest of the logic for additional dialog elements that you need :)

kevingilbert100:
Maybe this new feature request means building a more advanced router?

ericblade:
So, I haven't worked with the feature yet, *but* the basics of it seem to indicate that it provides you a way to get the Alexa service to prompt (with multiple prompts if necessary) for all the information required to trigger an Intent, before contacting your skill with it.

[...]

Issue #222, opened 2017-04-05
No definition in code for custom slots


ericblade:
(update: this is related to #114 )

One of the things that my team really likes about the alexa-app kit vs the official alexa-sdk, is that the schema and utterances are packaged right along with the code, so it's a rather attractive, declarative sort of style.

We just noticed as we were working out good ways to deploy things, that there is a lack of any kind of definition for custom slot data.  

This is something that I would be interested in working on, if the maintainers would be interested in it, but I'd also like to discuss what we would think would be the best way to handle doing it would be.

I think what we are thinking about doing, since Amazon is lacking any official way to update the "Interaction Model" page, is writing some code to automatically pull the schema, utterances, and slot definitions from the skill code, and then use curl or node or something to automatically push it to the Interaction Model page.  (or at least generate some text files that we can copy/paste from in the meantime)

So, I'm thinking that we could define slots as objects or strings (to keep compatibility), if object of form

```javascript
slots: {
    startDate: { type: 'AMAZON.DATE', },
    subject: { type: 'SUBJECTS', examples: [ 'discuss code', 'code design', 'implementation meeting', 'one-on-one meeting', 'all-hands meeting', 'meeting with boss', .... ] },
}
```

and then have slots() output something like:

```
slotName
(line-separated examples)
(blank line)
slotName2
(line-separated examples)
(blank line)
...
```

Also considering that we might want (a) function(s) that would return the schema, slots, and utterances as javascript objects, so that we can parse them directly from our build/deployment system.

What do the people with more experience with the app kit suggest?

dblock:
I am very interested in this - I have a similar problem in https://github.com/artsy/elderfield/ where I have a custom slot called `NAME` and a [set of instructions](https://github.com/artsy/elderfield/blob/master/DEPLOYMENT.md) that say _Add a custom slot type called NAME. Paste the contents of artists.txt and last_names.txt into the values box._. 

I think custom slots are reusable across the entire schema, so we need to find a way to declare them outside of a specific intent. So maybe we could leave the definition of slot inside an intent alone and make this simple a global thing?

```js
app.custom_slot('NAME', {
  examples: [ 'andy warhol', ... ]
});

app.custom_slot('NAME', {
  examples: function() { fs.readFileSync('artist_names.txt') 
});

app.intent('AboutIntent', {
    "slots": {
      "VALUE": "NAME"
    },
    "utterances": [
      "about {-|VALUE}"
    ]
  },
 ...
```

[...]

PR #268, opened 2017-08-12
Support for outputting Skill Builder Beta schema

lazerwalker:
Currently, people using this library to generate intent schemas and sample utterance lists cannot use the [Skill Builder Beta](https://developer.amazon.com/blogs/alexa/post/02d828b6-3144-46ea-9b4c-5ed2cbfadb9c/announcing-new-alexa-skill-builder-beta-a-tool-for-creating-skills), as the new schema JSON format is not compatible with the format being generated by `app.schema()`.

This adds an optional boolean argument to `app.schema()` that, if `true`, generates a schema that's readable by the new beta tool, as well as a `betaSchema` option to `app.express()` that gets threaded through for HTTP requests that result in a schema being generated/output.

For now, all this means is a few key names are different, there are some empty arrays, and each intent now includes its sample utterances inline. More importantly, this is necessary groundwork to eventually add library support for various new features afforded by the new tool, such as intent confirmation (https://github.com/alexa-js/alexa-app/issues/255), multi-turn dialogs, custom slot synonyms, and being able to declaratively specify custom slots in the schema itself  (https://github.com/alexa-js/alexa-app/issues/222).

Implementation wise, I intentionally kept this separate from the existing schema generation code — both the tests and the production code are just copy/pasted and modified versions of the existing schema code. We'll eventually want to make the new schema the default one, and perhaps even deprecate the old one, which suggests keeping them separate will make that easier. 

Sometime relatively soon I should have a corresponding PR in http://github.com/alexa-js/alexa-app-server that adds a configuration flag to output this in its default HTML template, but that shouldn't be a blocker to merging this in.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -63,10 +63,20 @@ export class app {
 
   request: (requestJSON: alexa.Request) => void;
 
-  /** Extracts the schema and generates a schema JSON object
-   * @param boolean useBeta If true, will return the new Skill Builder Beta schema
+  /** @deprecated It's recommended you directly call `app.schema.*` instead
+   * Extracts the schema and generates a schema JSON object
+   * This is equivalent to calling `app.schemas.intent()`
    */
-  schema: (useBeta: boolean) => string;
+  schema: () => string;
+
+  /** Functions to generate schema JSON objects in Amazon's various formats */
+  schemas: {
+    /** Generates a schema in the old default intent format */
+    intent(): string;
+
+    /** Generates a schema in the new Skill Builder beta format */
+    skillBuilder(): string;
+  };
 
   /** Generates a list of sample utterances */
   utterances: () => string;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -63,8 +63,10 @@ export class app {
 
   request: (requestJSON: alexa.Request) => void;
 
-  /** Extracts the schema and generates a schema JSON object */
-  schema: () => string;
+  /** Extracts the schema and generates a schema JSON object
+   * @param boolean useBeta If true, will return the new Skill Builder Beta schema
+   */
+  schema: (useBeta: boolean) => string;
 
   /** Generates a list of sample utterances */
   utterances: () => string;
@@ -82,6 +84,7 @@ export class app {
    * @param string options.endpoint the path to attach the router to (e.g., passing 'mine' attaches to '/mine')
    * @param bool options.checkCert when true, applies Alexa certificate checking (default true)
    * @param bool options.debug when true, sets up the route to handle GET requests (default false)
+   * @param bool options.betaSchema when true, Express debug requests will use the new Skill Builder Beta schema
    * @param function options.preRequest function to execute before every POST
    * @param function options.postRequest function to execute after every POST
    * @throws Error when router or expressApp options are not specified
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [#271](https://github.com/alexa-js/alexa-app/pull/271): Update request typing - [@jontg](https://github.com/jontg).
 * [#269](https://github.com/alexa-js/alexa-app/pull/269): Fix the lambda() type definition - [@jontg](https://github.com/jontg).
+* [#268](https://github.com/alexa-js/alexa-app/pull/268): Add support for outputting new Skill Builder schema - [@lazerwalker](https://github.com/lazerwalker).
 * Your contribution here
 
 ### 4.1.0 (August 12, 2017)
--- a/README.md
+++ b/README.md
@@ -513,7 +513,7 @@ app.post = function(request, response, type, exception) {
 
 ## Schema and Utterances
 
-The alexa-app module makes it easy to define your intent schema and generate many sample utterances. Optionally pass your schema definition along with your intent handler, and extract the generated content using the `schema()` and `utterances()` functions on your app.
+The alexa-app module makes it easy to define your intent schema and generate many sample utterances. Optionally pass your schema definition along with your intent handler, and extract the generated content using either the `schemas.intent()` and `utterances()` functions on your app (if using the normal Developer portal) or `schemas.skillBuilder()` if using the new Skill Builder beta.
 
 
 ### Schema Syntax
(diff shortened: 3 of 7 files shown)


dblock:
I just merged the other one, you should rebase.

dblock:
I am really not a big fan of this `if` and boolean beta parameters. This is not the first or the last time there's a beta schema and it will stop being beta at some point. 

Thinking out loud - maybe this is better?

* rename `betaSchema` to `skillBuilderSchema`, so we don't have to change it in the future
* add `options.schema` that defaults to the `schema` function, otherwise can be set to `skillBuilderSchema`, so that it's future proof for defining what generates the schema

What do you think?

lazerwalker:
If I'm understanding you correctly, `options.schema` will point to the actual function? So you'd e.g. 

```js
app.express({
    ..., 
    schema: app.skillBuilderSchema, // or a default of app.schema, etc
    ...
});
```

I think that sounds reasonable. Wonder if it'd make sense to namespace it one level more, so it's e.g. `app.schemas.skillBuilder`. 

Would also propose renaming the current `schema` to `intentSchema` or something else a bit more descriptive.

Also to think about: in a world where people are using the new skill builder schema, `app.utterances()` is no longer relevant. In a world where people are controlling which schema they're served up, it feels a little weird having it on the side there as a vestigial limb. I don't think there's necessarily anything actionable right now — I don't think a refactor that'd be a breaking change is worth it — but something to keep in mind.

ericblade:
Thanks, lazerwalker. 

I think I would want to be able to query the schema from the web interface, in either format.  It's possible that two different people can use the same code, one using skill builder, the other not.

lazerwalker:
@ericblade Are you talking about being able to access the exact same skill on the same local server and access different schemas, or are you talking about e.g. running multiple skills off of the same `alexa-app-server` instance and being able to have some skills use the old schema format and some of them use the new?

I agree the latter is valuable. I'd love to hear a bit more about the use case for wanting to see different schema representations for a single skill in the same local instance.

ericblade:
In my environment, we have two completely separate sets of virtually everything- a "production" and a "development".  We currently remain not using skillbuilder, but I'd like to switch over on the development side, so that I can make sure that everything works correctly.  So, I'd like to be able to query two different services, that are configured as alike as possible, with a parameter, to get whichever format of schema is needed.  

Though I think it might be possible to run multiple skills off multiple amazon accounts, using the exact same endpoints, but the accounts might be configured differently as well.  Not exactly my use case, but seems possible.  I'd want it to be a switch given in the request, rather than an initial code switch.

lazerwalker:
With what's described above, you could presumably have your schema output configuration vary based on an ENV variable that differs in your prod and development environments. 

Does that seem reasonable, or are you saying you need the exact same server instance to be able to spit out both schemes?

ericblade:
While that is certainly something that would work, I'm not seeing any particularly good reason to make it so that a server can output only one version of the schema.  I don't see any way that it would save any memory, execution time, load time, or developer time.  What advantage is there to making it so that it can only do one or the other?

Sure, once I migrate developer, make sure it all works right, then I migrate "production" (in quotes because we don't really have a production, but we do have an environment setup for it for when we do, "production" right now is "always ready for a product demonstration"), I'll probably never need it again.  But, say someone over in the corporate office on another continent wants to go set one up, they may not even have access to skillbuilder from where they are (didn't it start out as a very limited beta, that was then rolled out to most everyone?), and would have to make a code change.

dblock:
Lets start by making the schema configurable at runtime and make sure the parts that output schema can show both (maybe belongs in alexa-app-server?).

lazerwalker:
Yep, that sounds great. Sorry if my questioning came out more crotchety than intended :)

So right now, there are two ways the schema is exposed:

1. If you include a `schema` query param in a GET request, we send just the schema. I think this becomes a config flag, so `?schema=skillBuilder` gets you the new schema and anything else gets you the old.

2. If you don't include any query params, we render a template (which includes schema, utterances, etc), which is what `alexa-app-server` consumes. Presumably, we either want to include a different query param other than the already-in-use `schema` that toggles which schema gets included in this response, or just include both schema formats and let `alexa-app-server` decide which to consume. The latter's a bit less future-proof in a world where there may eventually be more than two schema formats. If we want the former, thoughts on a good name for that param?

dblock:
(1) sounds great

(2) I think we should be more future proof, either way the code in this library shouldn't be making the choice for alexa-app-server, it just needs to expose all functions

lazerwalker:
I could be misreading how alexa-app-server works, but based on some manual testing it appears that the test page is grabbing its schema data by hitting each app's exposed express endpoint, rather than directly calling the function. Does that sound right?

Unless we're going to change that, figuring out the shape of the data returned by that endpoint does seem like a decision that has to be made here rather than alexa-app-server.

dblock:
You're probably right, but I don't know what I am saying since I didn't dig deep into that. I would really try to think of alexa-app as a standalone thing, then worry about alexa-app-server.

lazerwalker:
Alright, next stab:

* `app.schemas.intent()` and `app.schemas.skillBuilder()` return their respective APIs
* `app.schema()` just returns the value of `app.schemas.intent()` for backwards compatibility

* `GET /endpoint` and `GET /endpoint?schema` return the intent schema by default, for backwards compatibility
* `GET /endpoint?schemaType=skillBuilder` and `GET /endpoint?schema&schemaType=skillBuilder` both include the skill builder schema (you can also e.g. manually pass in `schemaType=intent`

As a pleasant side effect, this works gracefully with alexa-app-server as-is. Default behavior is the old behavior, but if you add in the query param to your alexa-app-server URL it gets threaded through to the app properly and respected.

types/index.d.ts
```diff
@@ -63,8 +63,10 @@ export class app {
 
   request: (requestJSON: alexa.Request) => void;
 
-  /** Extracts the schema and generates a schema JSON object */
-  schema: () => string;
+  /** Extracts the schema and generates a schema JSON object
+   * @param boolean useBeta If true, will return the new Skill Builder Beta schema
+   */
+  schema: (useBeta: boolean) => string;
```
dblock: There's no longer a `useBeta` thing.

dblock:
This is great. Needs README, CHANGELOG, etc.

dblock:
I'm merging this.

PR #271, opened 2017-08-14
Update request session typing

jontg:
I'm reasonably certain the session in our `request` object is an `alexa.Session`...  Running locally I get all sorts of type warnings if I do what I believe is right
```bash
ERROR in ./handler.ts
(11,31): error TS2339: Property 'application' does not exist on type 'session'.
```

But on the other hand, if I try to follow the type definitions, I get
```bash
  Error --------------------------------------------------

  Unhandled exception: Cannot read property 'applicationId' of undefined.
```

On the *third* hand, if we circumvent the type definitions via (for example) the following code snippet, everything works hunky dory
```javascript
app.pre = (request, response, type) => {
  if (process.env.ALEXA_APP_ID !== undefined && !request.hasSession()) {
    // Ugly hack to work around Session type not being found properly
    const session : any = request.getSession();
    if ( session.application.applicationId != process.env.ALEXA_APP_ID) {
      console.log("Saw an unauthorized request from '%j'", request);
      throw "Invalid applicationId"; // Fail ungracefully
    }
  }
};
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.1.1 (Next)
 
+* [#271](https://github.com/alexa-js/alexa-app/pull/271): Update request typing - [@jontg](https://github.com/jontg).
 * Your contribution here
 
 ### 4.1.0 (August 12, 2017)
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -115,7 +115,7 @@ export class request {
   hasSession: () => boolean;
 
   /** Returns the session object */
-  getSession: () => session;
+  getSession: () => alexa.Session;
 
   /** Returns the request context */
   context?: alexa.Context;
--- a/types/test.ts
+++ b/types/test.ts
@@ -2,7 +2,16 @@ import * as alexa from 'alexa-app';
 
 const app = new alexa.app('app-name');
 
-app.pre = (request, response, type) => {};
+app.pre = (request, response, type) => {
+  if (request.applicationId !== undefined) {
+    // As documented in https://github.com/alexa-js/alexa-app#session
+    throw new Error('Invalid Application ID');
+  }
+  if (request.hasSession() && request.getSession().application.applicationId !== undefined) {
+    // As documented in https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/alexa-skills-kit-interface-reference#session-object
+    throw new Error('Invalid Application ID');
+  }
+};
 app.post = (request, response, type) => {};
 
 app.launch((request, response) => {});


types/index.d.ts
```diff
@@ -115,7 +115,7 @@ export class request {
   hasSession: () => boolean;
 
   /** Returns the session object */
-  getSession: () => session;
+  getSession: () => alexa.Session;
```
dblock: Needs a test please.

jontg:
@dblock happy to add tests; do you or @lazerwalker [since he authored the types to begin with](https://github.com/alexa-js/alexa-app/pull/266/files) have a preferred method or framework for testing typings?

I could add integration-class tests which attempt to compile and run a couple of basic `.ts` lambda's (which is how I discovered the type issues to begin with)...

lazerwalker:
@jontg There's already scaffolding for tests inside `types/test.ts`, which get compiled and checked for compiler errors on `npm run dtslint`, although I think you're right that for this specific case we'd actually need to run the code instead of just do a type-checking pass to catch/confirm this bug.

dblock:
Integration tests sounds about right! Would love to see tests move to a subfolder in `test`, too.

jontg:
Alright, I'll see if I can't PoC something tonight 🤔 

@dblock with regards to moving `test.ts` into the `test` sub-folder, [it looks like](https://github.com/Microsoft/dtslint) this is the recommended format - I tried briefly moving it over to the `test` top level folder and the tool does not pick this up.  _However_, we could create a sub-folder under `types/test/` if you would prefer that approach.

jontg:
Actually lets do this - @dblock I added the type test based on documentation (and provided the link), and I'll put something together for integration testing typescript in a separate PR.

lazerwalker:
Awesome, makes perfect sense to me.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### 4.1.1 (Next)
 
-* Your contribution here
```
dblock: Put this back please.

dblock:
So this fails without the fix? In what way? Just learning about TS ;)

lazerwalker:
Without the fix, the compiler/type-checker would e.g. see the code at https://github.com/alexa-js/alexa-app/pull/271/files#diff-160659ae3952dc97d42288b6145b5dccR10 and complain that the type definition for the object returned by the `getSession()` has no `application` property.

dblock:
Merged, thanks.

jontg:
```bash
$ node_modules/.bin/dtslint types/
/Users/jontg/Development/alexa-app/types/test.ts
ERROR: 10:52  expect  TypeScript compile error: Property 'application' does not exist on type 'session'.
```

jontg:
Woot, thank you @dblock and @lazerwalker for the feedback and fast responses!

lazerwalker:
Yeah, and thank you @jontg! Glad to see folks are already using the TS type definitions, and it's awesome you were willing to jump in and help — good to see some love given to some of the corners of the API's surface area that my own Alexa skill codebases didn't touch.

lazerwalker:
@jontg Looking back at this, I think this PR may be incorrect?

https://github.com/alexa-js/alexa-app/blob/master/index.js#L255-L261 shows `getSession()` returning a `session` object, rather than an Amazon-provided `Session` object (it's a bit confusing, since in the TS file "alexa.Session` refers to an Amazon type, whereas here `alexa.session` is our own object defined later in the file).

I'm surprised that `request.getSession().application.applicationId` is actually returning anything for you — running the latest copy of master locally, I'm seeing that property not exist, and reading through the code corroborates my expectations. (`request.getSession().details.application.applicationId` does exist).

If there's something I'm missing, any chance you could provide some code I could run to see where I'm going wrong?

(At a higher level, this points towards the value of getting some more integration-level tests in place)

tejashah88:
@dblock @lazerwalker Talking about the last point, I feel that that these current tests don't actually prove that the library is truly bug free or functionally ready. I'd propose keeping the tests we got so far into a separate folder and adding a new folder that tests the actual integration of alexa-app. This could also be a great place for hosting any example skills as part of the documentation.

There's a node module called [chatskills](https://github.com/primaryobjects/chatskills) that allows text-based emulation of any alexa skill, although it's specifically targeted for skills using this library. We could use that as a basis for testing out the "functional" features.

Another option could be either making test skills on AWS lambda and Heroku (or any other web service) to test the said functionality, or being able to call the skill directly (via [node-lambda](https://github.com/motdotla/node-lambda) or simple POST requests).

dblock:
My philosophy is to keep adding tests until they start proving something. And either way write regression tests every time. So PR along. 

I don't think we should be running integration tests though against any kind of production like environment because that makes it hard for developers to setup. Node-lambda though is a good idea as long as it's self-contained.

lazerwalker:
It seems the debug web page on `alexa-app-server` is already capable of generating well-formed request payloads to the server in the same format of actual Amazon request payloads, yeah? I haven't looked into how that works, but using that infrastructure as the basis of some simple tests that generate requests and assert on the response payload seems sensible to me.

PR #275, opened 2017-08-18
Add support for custom slot types 

### LINKED ISSUE #114: Custom Slot Types don't appear in Alexa Tester
Not sure what I'm missing and if this is a problem with alexa-app or alexa-app-server.

Trying to use a custom slot type, which basically looks like this in my intent:

```
    {
        "slots":{"COMMAND":"COMMAND_SLOT"},
        "utterances":[ "send {the|} {command|} {commands|COMMAND}" ]
    }
```
(where commands is an app.dictionary)

I've also tried:

```
    {
        "slots":{"COMMAND":"COMMAND_SLOT"},
        "utterances":[ "send {the|} {command|} {-|COMMAND}" ]
    }
```

In both cases, the Utterances properly appear in the Alexa Tester (using alexa-app-server 2.3.1), but the Custom Slot Type section of the tester is always empty. I was under the impression the custom slot type info should appear there to facilitate copy/pasting into the interaction model.

I started on alexa-app 2.3.4 and upgraded to 2.4.0 ... same problem.

Thanks!

lazerwalker:
This adds in support for defining your own custom slot types.

The use case for this: I have a script that automatically takes my generated Skill Builder schema, navigates to the Amazon dev portal, and uploads the new schema file + rebuilds the interaction model. For my skills that have custom slot types, this means uploading a new schema file overwrites my existing custom slot types, which renders that workflow useless. Adding first-class support for custom slot types within the library itself means that the Skill Builder schema file it generates can now contain 100% of your interaction model for skills with custom slot types.

 In its simplest form, you can add a custom slot type via `app.customSlot("SlotName", ["value1", "value2", "value3"])`. There's a slightly more verbose syntax if you want to take advantage of the new Skill Builder features, like synonyms.

If you do that, and then call `app.schemas.skillBuilder()`, the resulting schema object contains your custom slots, ready to paste into the Skill Builder UI.

This doesn't affect intents or actually handling requests (nor should there really be any reason to). It exists purely to thread the data through to the schema JSON.

This would resolve #114 and #222.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 * [#268](https://github.com/alexa-js/alexa-app/pull/268): Add support for outputting new Skill Builder schema - [@lazerwalker](https://github.com/lazerwalker).
 * [#273](https://github.com/alexa-js/alexa-app/pull/273): Correct inconsistent slots API documentation and interface definitions - [@padraigkitterick](https://github.com/padraigkitterick).
 * [#274](https://github.com/alexa-js/alexa-app/pull/274): Correct inconsistent interface definitions for optional ExpressOptions properties - [@lazerwalker](https://github.com/lazerwalker).
+* [#275](https://github.com/alexa-js/alexa-app/pull/275): Add support for custom slot types (for Skill Builder schema output) - [@lazerwalker](https://github.com/lazerwalker).
 * Your contribution here
 
 ### 4.1.0 (August 12, 2017)
--- a/README.md
+++ b/README.md
@@ -564,6 +564,24 @@ sampleIntent     airport status for {CustomSlotName}
 
 Note that the "CustomSlotType" type values must be specified in the Skill Interface's Interaction Model for the custom slot type to function correctly.
 
+#### custom slot type values
+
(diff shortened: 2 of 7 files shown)


dblock:
This is perfect, merging!

jradwan:
Nice! So this will be part of the next 4.1.1 release?

lazerwalker:
Yep, it'll be part of the next release (which I suspect should actually be called `4.2.0`.

Unless anyone else feels differently, my plan was to wait until a few other pending threads were closed before cutting a new release. Tracking here: https://github.com/alexa-js/alexa-app/issues/277

lazerwalker:
@jradwan I just pushed a new version `4.2.0` that has these changes, if that's of use to you. 🎉

jradwan:
Great, thanks @lazerwalker! I have no internet at home (after Hurricane Irma - power, but no TV/internet/phone) but once I do I'm looking forward to testing this out. 😄

lazerwalker:
Oh no! Hope things aren’t too bad :(

jradwan:
We were lucky ... no major damage (lost a fence, lots of tree debris), got power back after 21 hours and internet after 5 days. Thanks for asking!

jradwan:
Gave this a shot and see how the app.schema.skillBuilder call works, nice. Now if I could just get the results into the Alexa Tester UI for the alexa-app-server.

Thanks again!

PR #278, opened 2017-08-20
Rewrite tests in TS

lazerwalker:
We've been running into a number of issues with the new TS definitions not matching up with production code, including a case where a PR to change the TS definitions ended up introducing actively wrong types (https://github.com/alexa-js/alexa-app/pull/271).

The solution here is clearly better tests that actually ensure that the TS types reflect reality.

One way to do that is to write a whole new test suite that extensively touches every corner of the codebase, and is also written in TS. That sounds like a lot of work, and seems like something that's liable to fall out-of-sync pretty quickly with the mainline test suite. A perhaps more pragmatic way to go about that is to rewrite the existing tests to be in TS.

Here's a proof-of-concept of rewriting a single existing spec file into TS. It's not yet as much of a functioning developer flow as I'd like (before merging, I would make sure that running `npm run test` handles everything behind the scenes), but even the act of doing this single file drove out a bunch of corrections to both `index.d.ts` and `alexa.d.ts` (including, to be explicit, failing tests surrounding the issues in #271), proving there's value here. 

Before I go through and convert more files, I wanted to gauge responses to this solution. 

The main drawback is that anyone who wants to write/modify tests now has to write code in TS instead of JS, and deal with the TS compiler. It seems like there's historically been [some opposition](https://github.com/alexa-js/alexa-app/issues/148) to that in this project. Tooling changes would be made completely behind-the-scenes (i.e. running `npm run test` would run the entire test suite, as it does now), and you'll see the test file itself has relatively minimal changes.

This does still leave the production code itself in vanilla JS, and lets us maintain a single test suite going forward.


lazerwalker:
(Coveralls coverage decreasing is expected. If this is something we want to follow through on, presumably tweaking the coveralls config to include .ts tests is doable/simple)

dblock:
I have mixed feelings about this. The next step will be to rewrite alexa-app in TS and we had talked about this before and decided it was a bad idea. I'd like to hear other opinions.

lazerwalker:
Yeah, I have equally mixed feelings.

I think what's become clear is that, in order for the TS type definitions to be accurate, there needs to be a sample app that touches all features of the library, written in TS and passing TS compilation, and with some level of proof that the app itself actually behaves as expected.

I want to spend a little more time banging on the tests to see if I can get `tsc` to successfully type-check the existing JS files with little or no changes (including, critically, leaving them as JS instead of TS). I'm not sure I can (I think there may be too much type ambiguity), but if so that might be the perfect compromise.

lazerwalker:
Alright, next stab at this.

Here are a few test files, still in JS, but now properly type-checking via `tsc` with only minor changes needed. A few things still are not great:

* There are a few cases (in `test_alexa_app_session.js`) where type ambiguity meant I needed to insert type annotations via JSDoc (who knew TypeScript could do that!). In the few files I did, this was the only place I really needed to do that.
* TypeScript requires ES6 `import` statements rather than CommonJS `require`, which necessitates pulling in Babel solely as a dev dependency so mocha can run the test files. Given we've already bloated devDependencies with typescript itself, this doesn't feel too bad to me.

Those few unfortunate things aside, this does leave us with our tests in JS, while also allowing us to validate the TypeScript type definitions against actual validated code without worrying about a separate test suite falling out-of-sync. 

Thoughts?

dblock:
This looks innocent enough. As long as tests continue being in JS and it's not insane for a developer to add more tests I am OK with this.

lazerwalker:
FYI, the work for this is essentially done, but seeing CI go green is blocked on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/19382.

lazerwalker:
@dblock Wanna take a look at this? Think it should be good to go — tests are passing, both the test suite itself and the type-checking pass.

On `npm run dtslint`, the type checker runs on every JS test file except for the two that are testing smaller standalone helper functions (`normalize_api_path` and `ssml`) and the express and lambda integrations (they were significantly more complex to make typescript-happy, and the lambda ones in particular seem like they may actually be false positives?).

CONTRIBUTING.md
```diff
@@ -58,6 +58,8 @@ npm install
 npm run dtslint
 ```
 
+This will also run the TypeScript type checker on all of the unit tests (even though they're in JavaScript!) to make sure that reality matches the type definitions. This may fail, possibly with an error that there's some type ambiguity, which might mean you need to tweak the tests a bit to give TypeScript the info it needs. If you're having trouble with this, you may want to look at other tests in this project to see workarounds for similar problems, or barring that feel free to bug [@lazerwalker](https://github.com/lazerwalker).
```
dblock: Thanks for sticking your head out here, although you don't have to :)

dblock:
I like this enough. Lets give it a shot, merging.

CONTRIBUTING.md
lazerwalker: Yeah, definitely ready to erase that line if it turns out to be a monumentally terrible decision :P.

Issue #279, opened 2017-09-02
Wrong Typescript definition of Request, Session and User


BeneStem:
Hey guys,

the current master branch has a wrong typescript definition of request, session and user object.

They should be:
```
getSession: () => {
    details: alexa.Session,
    sessionId: string,
    attributes: any
  };
```
```
export interface Session {
  new: boolean;
  sessionId: string;
  attributes: any;
  application: {
    applicationId: string;
  };
  user: User;
  userId: string;
  accessToken: string | null;
}
```
```
export interface User {
    userId: string;
    accessToken?: string;
}
```

Here is some evidence from cloudwatch:
```
console.log(request.getSession())
{
  isAvailable: [
    Function
  ],
  isNew: [
    Function
  ],
  get: [
    Function
  ],
  set: [
    Function
  ],
  clear: [
    Function
  ],
  details: {
    new: true,
    sessionId: 'YOU WISH :P',
    application: {
      applicationId: 'YOU WISH :P'
    },
    user: {
      userId: 'YOU WISH :P',
      accessToken: 'YOU WISH :P'
    },
    userId: 'YOU WISH :P',
    accessToken: 'YOU WISH :P'
  },
  attributes: {},
  sessionId: 'YOU WISH :P',
  getAttributes: [
    Function
  ]
}
```

Would be nice if you could fix it.
Greetings
/ben

dblock:
Would be nice if you could PR a fix, please?

BeneStem:
sure. will do that.

PR #280, opened 2017-09-02
Fix typedefinitions of request, session and user

BeneStem:
Fixes: #279

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.1.1 (Next)
 
+* [#280](https://github.com/alexa-js/alexa-app/pull/280): Fix session typing - [@benestem](https://github.com/benestem).
 * [#271](https://github.com/alexa-js/alexa-app/pull/271): Update request typing - [@jontg](https://github.com/jontg).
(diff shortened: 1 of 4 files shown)


BeneStem:
Strange. Test says that "TypeScript compile error: Property 'application' does not exist on type '{ details: Session; sessionId: string; attributes: any; }'."
But in the real world it does :-/
What am I missing?

BeneStem:
Updated the commit with force push to squash it... Not sure if this was the correct way to do it :/

dblock:
Something related to the version of node? Maybe @lazerwalker can help?

lazerwalker:
That's an expected `dtslint` failure — the existing TS tests essentially had an assertion on the old signature that `getSession()` returned an `alexa.Session` instead of a `session`.

Taking [this line](https://github.com/alexa-js/alexa-app/blob/master/types/test.ts#L10) and swapping in `request.getSession().application.applicationId` should fix it.

I think that + a changelog entry should have this PR good to go?

Thanks!

BeneStem:
Ok. Will do that. Gimme a sec.

BeneStem:
ok. done.

BeneStem:
Thank you for being so patient and helpful :)

BeneStem:
it's green. :P

dblock:
@lazerwalker This is correct? Please merge.

PR #284, opened 2017-09-14
Add schema support for ask-cli

lazerwalker:
Amazon recently released a [CLI](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/ask-cli-intro) to (among other things) upload changes to your interaction model.

This PR adds a new `app.schemas.askcli` function that can generate JSON in the correct format, such that it will work with `ask deploy` as-is if you write it to disk in the correct folder structure. It's basically the Skill Builder schema, but (a) the top-level wrapping structure is sliiiightly different and (b) it includes the skill's invocation name.

To the latter point: there's now an `app.invocationName` property that can be set, with it falling back to `app.name` if there isn't an invocation name set. The `app.schemas.askcli()` function can also take in an invocation name. I personally have staging and production deployments of the same skill with slightly different invocation names (I assume this is a fairly common use case), so being able to specify that at schema generation time is useful.

I suspect the tooling work I'm doing in my own skill may eventually be able to be abstracted out into a tighter-knit integration, but for now this feels like the right level of adaptation to this library to facilitate working with `ask-cli`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ### 4.2.1 (next)
 
+* [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
 * Your contribution here
+
 
 ### 4.2.0 (September 13, 2017)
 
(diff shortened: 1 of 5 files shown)


dblock:
LGTM, needs CHANGELOG

matt-kruse:
Will this feature be released soon?

dblock:
Just to close the loop, thanks for opening #309.

PR #286, opened 2017-09-19
If the user says a custom slot synonym, request.slot() returns the main value instead of the spoken synonym

lazerwalker:
When defining a custom slot, Amazon now allows you to define synonyms for given slot values (something our `app.customSlot` syntax supports).

When a user says a synonym instead of the main custom slot value, the value returned within the intent request (which `alexa-app` in turn picks up for e.g. `request.slot()`) is the synonym rather than the actual value.

To put this in more concrete terms: say you have a custom slot value of "cat" with a synonym of "kitty". If the user says "kitty", the intent will come through as if the user said "kitty" (e.g. `request.slot('pet') == "kitty"`). As a developer, I now manually need to say "okay, I know that 'kitty' isn't in the list of things I was expecting, let's do a lookup of what it's a synonym of". 

It seems more intuitive to me that we'd automatically coerce synonyms into their known main values. This PR makes it so that if the user says "kitty" for a slot called "pet", `request.slot('pet')` (or `request.slots.pet.value`) would return "cat" rather than "kitty". The original "kitty" string is now available at `request.slots.pet.rawValue`.

This is an opinionated change, but it feels more in-line with how I'd intuitively expect "synonyms" to behave in this context.

Thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.2.1 (next)
 
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
+* [#286](https://github.com/alexa-js/alexa-app/pull/286): When an intent request contains a custom slot value synonym, coerce the slot value into the main value instead of the synonym - [@lazerwalker](https://github.com/lazerwalker).
 * Your contribution here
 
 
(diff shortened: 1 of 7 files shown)


dblock:
For now just thinking out loud - It feels to me like we may be trying to be "too smart" here. Will this be hard to undo if the resolution becomes more complicated? I'm thinking translations will come next, or multiple possible resolutions in priority order. A first class helper for resolutions may be a better idea rather than hiding the entire thing to the developer?

lazerwalker:
Closing this out for now. If others had felt this pain, I miiight want to consider this specific PR a bit more. Otherwise, my instinct is yeah, we want a first class resolutions helper, but that probably shouldn't be designed until more than 1 resolution type actually exists.

PR #292, opened 2017-10-11
Adding documentation to specifically call out support for the Echo Show.

jstreb:


jstreb:
This is just updates to the readme.md file. Not sure why the build is failing.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 4.2.1 (next)
 
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
+* [#292](https://github.com/alexa-js/alexa-app/pull/292): Updating README.md with documentation regarding echo show support. = [@jstreb]
```
dblock: You should remove this. It doesn't match the format, but other than that we don't CHANGELOG README changes.

README.md
```diff
@@ -34,6 +34,7 @@
 * [Custom Directives](#custom-directives)
 * [Dialog](#dialog)
 * [Error Handling](#error-handling)
+* [Show Support](#show-support)
```
dblock: It's an "Echo Show" right? I would say that.

README.md
```diff
@@ -891,6 +892,21 @@ app.error = function(exception, request, response) {
...
+```
+app.post(req, res, type, exception) {  
+  // If the device does not support display directives then remove them from the response
+  if (!system.supportsDisplay(req))) {
+    res.response.response.directives = []
+  }
+}
+```
+
+Please reference [Amazon's documentation](https://developer.amazon.com/docs/custom-skills/display-interface-reference.html#display-template-reference) for the list of supported template markup.
```
dblock: "refer to", not "reference" :)

dblock:
Thanks for doing this! Some minor comments.

dblock:
Looks like our build is broken on node 1.0, maybe @lazerwalker or someone else can help.

CHANGELOG.md
jstreb: Perfect. I only added it because of warning in the build.

README.md
jstreb: 👍

README.md
jstreb: true

PR #294, opened 2017-10-12
WIP: Attempt to fix Travis

lazerwalker:
Our Travis is currently failing for two reasons:

1. The latest version of dtslint requires that we set a tsconfig.json option that we were not previously setting.
2. A sub-sub-sub-dependency (hawk) has pushed an update that fails on Node v1.0

Both of these (the second more than the first) are presumably(?) caused by the fact that we weren't pegging down version requirements in a stricter way than the default package.json file.

This commit does two things:

1. Adds in the `strictFunctionTypes` setting to tsconfig.json for future-proofing
2. Commits npm's `package-lock.json` to ensure both Travis and all users are running the exact same versions of dependencies.

I personally have a slight preference for yarn over npm, but figured adding npm's lock solution was probably lower-friction.

Dunno if this will work yet, but I'm not sure Travis is set up right now to run on branches on my fork, so opening this WIP PR is easier :)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,3618 @@
+{
+  "name": "alexa-app",
+  "version": "4.2.1",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@types/chai": {
+      "version": "4.0.4",
+      "resolved": "https://registry.npmjs.org/@types/chai/-/chai-4.0.4.tgz",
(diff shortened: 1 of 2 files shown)


lazerwalker:
Cool, looks like the dtslint change worked, but need to investigate what's going on with package-lock.json

lazerwalker:
Made irrelevant by #301, which includes my `strictFunctionTypes` fix and the underlying node 1.0 issue.

PR #301, opened 2017-11-20
updated alexa-verifier-middleware to v1.0.0

tejashah88:
There has been many underlying changes to [alexa-verifier](https://github.com/mreinstein/alexa-verifier), particularly with improvements to the stability and security of the module. Recently it's reached v1.0.0, so I have thus updated the middleware module to reflect the recent changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.2.1 (next)
 
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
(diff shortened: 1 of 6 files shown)


tejashah88:
Seems like there's a problem with node v9.2.0 and typescript? @lazerwalker 

image

lazerwalker:
Just pushed a fix to this branch that should fix CI. TypeScript frustratingly changed things so that a formerly-optional config key is now mandatory, so adding it in should fix things.

Otherwise, this looks 👍 to me!

lazerwalker:
Fixed up the various other tslint issues that arose as a result of bumping versions.

The only slightly sketchy one is that we seem to be running into a dtslint/compiler bug that means it was incorrectly yelling at us for what should have been a valid TS definition. Had to manually add in a linter config to ignore that rule while I open an issue with the actual dtslint repo.

Will leave merging to you, but I'm satisfied? Your stuff looks good, and CI's green.

tejashah88:
Thanks for helping out with the typescript shenanigans!

PR #302, opened 2017-11-22
Move optionalDependencies to devDependencies

### LINKED ISSUE #300: Switch optionalDependencies to peerDependencies
### Description

This package currently has `typescript` and `dtslint` listed as `optionalDependencies`.

I do not use typescript in my lambda functions and have no intention of doing so. 

However because this package was listed as an `optionalDependency` rather then a `peerDependency` its causing me to zip it up with the rest of my dependencies. (also increasing my artifact by ~10MB)

### Reference

https://github.com/alexa-js/alexa-app/blob/a32fc071d6fa6baba48b9cdf3409bb0c58c85e1c/package.json#L63-L66

lazerwalker:
We have typescript and dtslint set up as `optionalDependencies` in `package.json`. This means they get installed when this is pulled in as a client lib, even though people consuming this don't need typescript. This moves them to `devDependencies` instead, which should mean they'll only be installed when someone explicitly clones this repo and runs `npm install` to get a dev setup!

Fixes #300.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
 * [#301](https://github.com/alexa-js/alexa-app/pull/301): Updated alexa-verifier-middleware to v1.0.0 - [@tejashah88](https://github.com/tejashah88).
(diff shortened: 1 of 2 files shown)


lazerwalker:
Argh, looks like this'll go green once it's rebased against #301. Don't think there's a huge rush, so I'll revisit this once that hits master.

lazerwalker:
Cool, seems to be passing now 👍

Issue #309, opened 2018-01-26
Release 4.2.1


matt-kruse:
Please release 4.2.1, so new projects can use the askcli() feature. Thanks.

dblock:
You're the OG on this project Matt and you have all access to do a release! Go for it.

dblock:
Updated release instructions in https://github.com/alexa-js/alexa-app/blob/master/RELEASING.md

dblock:
@kobim Care to do a release? Follow the instructions above please.

matt-kruse:
I need to catch up a bit to make sure that I follow the procedures you have in place now. I'm not used to adjusting type definitions or test cases. ;) 

I'm working off a local copy now and make a few changes to support a new API that is coming. I won't be able to commit until the API is public, but if I get chance to do a release before then I will.

kobim:
@matt-kruse - anything you want to add to this release, or can I go ahead and publish it?

dblock:
Go ahead @kobim!

kobim:
Done!

matt-kruse:
Thank you! I am adding code for upcoming API's but didn't have anything to contribute for this release. But having this update on npm will be helpful, thanks!

PR #310, opened 2018-01-27
Feature: Slot Resolutions

kobim:
Each slot can allow _resolutions_ for the value, allowing developers to process values like "the biggest airport" which translate to "JFK" (and a possible pre-defined id).

This API allows accessing the slots resolutions (if any), making it easier to define synonyms and custom IDs in the interaction model.

Reference: https://developer.amazon.com/docs/custom-skills/define-synonyms-and-ids-for-slot-type-values-entity-resolution.html
Type definition: https://developer.amazon.com/docs/custom-skills/request-types-reference.html#resolutions-object

** Small note: the express tests were failing due a [recent change in the alexa-verifier-middleware](https://github.com/alexa-js/alexa-verifier-middleware/commit/38ec0c37b15c75ff93cbb650a27c4dda48b62457), hence the change in _test_alexa_integration_express.js_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
 * [#301](https://github.com/alexa-js/alexa-app/pull/301): Updated alexa-verifier-middleware to v1.0.0 - [@tejashah88](https://github.com/tejashah88).
 * [#302](https://github.com/alexa-js/alexa-app/pull/302): Move typescript and dtslint from optionalDependencies to devDependencies - [@lazerwalker](https://github.com/lazerwalker).
-* [#310](https://github.com/alexa-js/alexa-app/pull/310): Feature: Slot Resolutions - [@kobim](https://github.com/kobim)
+* [#310](https://github.com/alexa-js/alexa-app/pull/310): Added support for slot resolutions - [@kobim](https://github.com/kobim).
 * Your contribution here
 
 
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
 * [#301](https://github.com/alexa-js/alexa-app/pull/301): Updated alexa-verifier-middleware to v1.0.0 - [@tejashah88](https://github.com/tejashah88).
 * [#302](https://github.com/alexa-js/alexa-app/pull/302): Move typescript and dtslint from optionalDependencies to devDependencies - [@lazerwalker](https://github.com/lazerwalker).
+* [#310](https://github.com/alexa-js/alexa-app/pull/310): Feature: Slot Resolutions - [@kobim](https://github.com/kobim)
```
dblock: Should just say "Added support for slot resolutions" and is missing a period at the end of the line ;)

test/test_alexa_integration_express.js
```diff
@@ -211,7 +208,7 @@ describe("Alexa", function() {
       it("requires a cert header", function() {
         return request(testServer)
           .post('/testApp')
-          .expect(401).then(function(res) {
+          .expect(400).then(function(res) {
```
dblock: I know this is unrelated, but can we please track this one down and make sure that's not a bug?

kobim: I checked the alexa-verifier-middleware and for this change [(38ec0c)](https://github.com/alexa-js/alexa-verifier-middleware/commit/38ec0c37b15c75ff93cbb650a27c4dda48b62457). There was no explanation for this change, but they also [fixed their tests](https://github.com/alexa-js/alexa-app/pull/310#pullrequestreview-92028980) and released v1.0.1 with this change.

dblock:
@kobim You should be able to do the release, I added you to npm, see https://github.com/alexa-js/alexa-app/issues/309 please.

PR #314, opened 2018-02-10
Fix breaking logic in request.slot()

User1m:
I started receiving the following Intent Requests when an empty slot is provided:
```
{
    "type": "IntentRequest",
    "requestId": "amzn1.echo-api.request.ID",
    "timestamp": "2018-02-10T18:35:41Z",
    "locale": "en-US",
    "intent": {
        "name": "PersonIntent",
        "confirmationStatus": "NONE",
        "slots": {
            "Name": {
                "name": "Name",
                "confirmationStatus": "NONE"
            }
        }
    }
}
```
This was causing my function `request.slot(SlotName, defaultValue)` to fail. Because as it stands the logic in `request.slot()` doesn't check if the `value` is `'undefined'` and so doesn't return the `'defaultValue'`. This fixes that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.2.2 (Next)
 
+* [#315](https://github.com/alexa-js/alexa-app/pull/315): Fix card tests and required card type - [@kobim](https://github.com/kobim).
 * Your contribution here.
 * [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot(). - [@User1m](https://github.com/user1m)
 
@@ -11,6 +12,7 @@
 * [#301](https://github.com/alexa-js/alexa-app/pull/301): Updated alexa-verifier-middleware to v1.0.0 - [@tejashah88](https://github.com/tejashah88).
 * [#302](https://github.com/alexa-js/alexa-app/pull/302): Move typescript and dtslint from optionalDependencies to devDependencies - [@lazerwalker](https://github.com/lazerwalker).
 * [#310](https://github.com/alexa-js/alexa-app/pull/310): Added support for slot resolutions - [@kobim](https://github.com/kobim).
+* [#316](https://github.com/alexa-js/alexa-app/pull/3107): Fix dtslint failures - [@lazerwalker](https://github.com/lazerwalker)
 
 ### 4.2.0 (September 13, 2017)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 4.2.2 (Next)
+
+* Your contribution here.
+
 ### 4.2.1 (February 1, 2018)
 
 * [#284](https://github.com/alexa-js/alexa-app/pull/284): Add support for generating `ask-cli`-compatible schema JSON - [@lazerwalker](https://github.com/lazerwalker).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### 4.2.2 (Next)
 
 * [#315](https://github.com/alexa-js/alexa-app/pull/315): Fix card tests and required card type - [@kobim](https://github.com/kobim).
-* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot() - [@User1m](https://github.com/user1m).
+* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix handling of `undefined` slot values - [@User1m](https://github.com/user1m).
 * Your contribution here.
 
 ### 4.2.1 (February 1, 2018)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ### 4.2.2 (Next)
 
 * Your contribution here.
-* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot() - [@User1m](https://github.com/user1m)
+* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot(). - [@User1m](https://github.com/user1m)
 
 ### 4.2.1 (February 1, 2018)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 4.2.2 (Next)
 
 * Your contribution here.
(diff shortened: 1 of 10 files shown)


dblock:
Thanks. This needs tests and changelog, please.

User1m:
will update soon 👍

User1m:
@dblock Updates made. All but 1 tests pass due to a dslint error in `types/index.d.ts` which I did not modify. Would look into it more  if you want me to.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 4.2.2 (Next)
 
 * Your contribution here.
+* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot() - [@User1m](https://github.com/user1m)
```
dblock: My ADD says this is missing a period at the end ;)

test/test_alexa_app_customslot.js
```diff
@@ -10,6 +10,30 @@ chai.config.includeStack = true;
...
+      testReq = null;
+    });
+
+    it("get slot value if correct", function () {
+      var req = new Alexa.request(testReq);
+      expect(req.slot("AirportCode", "SEA")).to.eql("JFK");
+    });
+
+    it("gets defaultValue if slot value is undefined", function () {
+      testReq.request.intent.slots.AirportCode.value = undefined;
```
kobim: If you make sure a change, note that `mockRequest` itself is modified (as Objects are by-ref), and if a new test will be written after this it will fail.
I suggest loading the JSON in the `beforeEach`, to ensure there will be no affect on other test cases.

kobim:
@dblock there's some problem with dtslint's _strict-export-declare-modifiers_ rule ([ref 1](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/22503#issuecomment-354200016), [ref 2](https://github.com/Microsoft/dtslint/issues/88)).
Maybe it's worth disabling for now? (it affects [other builds as well](https://travis-ci.org/alexa-js/alexa-app/jobs/337861983))

test/test_alexa_app_customslot.js
User1m: good point. updated.

CHANGELOG.md
User1m: sounds good.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 4.2.2 (Next)
 
 * Your contribution here.
+* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot(). - [@User1m](https://github.com/user1m)
```
dblock: More OCD: The period after `user1m).`, check the other lines :)

lazerwalker:
Good news / bad news: on a hunch, I had Travis re-build master, and it's now red.

The problem is we're pointing dtslint at its `production` branch, as per https://github.com/Microsoft/dtslint/issues/71. 

I'm going to try switching to `0.2.0` in another branch, and re-open an issue in the dtslint repo for them to figure their versioning shit out.

lazerwalker:
Yeah, doing some IT work on our dtslint config cleaned that up (albeit in the form of ignoring the error, rather than switching versions). See #316. 

@User1m want to rebase off of that branch?

CHANGELOG.md
User1m: done haha sorry should've been more observant. thanks!

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### 4.2.2 (Next)
 
 * [#315](https://github.com/alexa-js/alexa-app/pull/315): Fix card tests and required card type - [@kobim](https://github.com/kobim).
+* [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix breaking logic in request.slot() - [@User1m](https://github.com/user1m).
```
User1m: ping @dblock

User1m:
Thanks @lazerwalker  for the fix 👍 
@dblock we're all green. Would love for a merge soon.
Thanks guys

test/test_alexa_app_customslot.js
```diff
@@ -12,7 +12,8 @@ import * as Alexa from '..';
 describe("Alexa", function() {
 
   context("#alexa.request.slot", function () {
-    var testReq;
+    var testReq = mockHelper.load("intent_request_airport_info.json");
```
kobim: Any reason you assigned `testReq` in the context? the previous commit was perfect.

test/test_alexa_app_customslot.js
```diff
@@ -10,6 +10,30 @@ chai.config.includeStack = true;
 import * as Alexa from '..';
 
 describe("Alexa", function() {
+
+  context("#alexa.request.slot", function () {
+    var testReq = mockHelper.load("intent_request_airport_info.json");
+    
+    beforeEach(function () {
+      testReq = mockHelper.load("intent_request_airport_info.json");
```
dblock: Seems like we're doing this twice, here and above?

CHANGELOG.md
dblock: Much better! ;) I would prefer that it actually explains the problem. From the code it looks like a slot value is set to `undefined` instead of not being present? Maybe "Fix handling of undefined slot values"? No strong feelings though.

test/test_alexa_app_customslot.js
```diff
@@ -10,6 +10,30 @@ chai.config.includeStack = true;
...
+
+    it("gets defaultValue if slot value is undefined", function () {
+      testReq.request.intent.slots.AirportCode.value = undefined;
+      var req = new Alexa.request(testReq);
+      expect(req.slot("AirportCode", "SEA")).to.eql("SEA");
+    });
+  });
+  
   describe("app", function() {
     var testApp = new Alexa.app("testApp");
```
User1m: @dblock was just following the convention below. It seems this is a work-around for `dtslint` causing issues if you just declare a variable placeholder say `var testApp` without assigning it.

User1m: @dblock - reference to your comment on line 18

User1m: by just declaring ` var testReq;` ts rightfully complains `Variable 'testReq' implicitly has an 'any' type.`. But in a JS file you cant declare a type to silence the error

User1m: I would advocate for changing `"noImplicitAny": true,` to `false` inside the `test/tsconfig.json` file

dblock: Feel free to PR separately whatever you see fit, we're splitting hairs ;)

dblock:
Merged, thanks for hanging in here with my OCD!

PR #315, opened 2018-02-19
Fix: card tests weren't failing the suite

kobim:
Tests were completing successfully although 1 case were failing because `eventually` assertions should be returned in the test case:
screen shot 2018-02-19 at 10 49 50
(committing the fix to the test first, to make sure the fix works and tests fail.)

As for the fix itself, the [Card object](https://developer.amazon.com/docs/custom-skills/request-and-response-json-reference.html#card-object) states the `type` is required, hence my change added it as a required field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### 4.2.2 (Next)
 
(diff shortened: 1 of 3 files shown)


kobim:
Synced with master.

dblock:
Merged, thank you.

PR #319, opened 2018-03-15
SkillBuilder Example

funkydan2:
Added an example of how to output the JSON required for the new SkillBuilder.

I'm new to this, so I'm not sure if this is the right & best way, but it works for me.


dblock:
Do update changelog please? Maybe we need to say something about this in README?

funkydan2:
Trying in another PR (sorry, I'm a github newby).

PR #324, opened 2018-03-24
Added TypeScript declarations for the Alexa address objects

CodeCutterUK:
Added TypeScript declarations for responses returned from the device-api

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * [#315](https://github.com/alexa-js/alexa-app/pull/315): Fix card tests and required card type - [@kobim](https://github.com/kobim).
 * [#314](https://github.com/alexa-js/alexa-app/pull/314): Fix handling of `undefined` slot values - [@User1m](https://github.com/user1m).
 * [#322](https://github.com/alexa-js/alexa-app/pull/322): Included Skill Builder example in example.ejs - [@funkydan2](https://github.com/funkydan2).
+* [#324](https://github.com/alexa-js/alexa-app/pull/324): Added TypeScript declarations for the Alexa address objects - [@CodeCutterUK)](https://github.com/CodeCutterUK).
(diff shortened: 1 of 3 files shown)


CodeCutterUK:
Looks like builds for node 1 to 4 are failing, but I don't see how my commit would have caused this :)

kobim:
The failure actually relates to npm v2, which tries to satisfies the _peerDependency_ `"typescript": ">=2.1.0 || >=2.1.0-dev || >=2.2.0-dev || >=2.3.0-dev || >=2.4.0-dev || >=2.5.0-dev || >=2.6.0-dev || >=2.7.0-dev || >=2.8.0-dev"` from [tslint](https://github.com/palantir/tslint/), but [dtslint](https://github.com/Microsoft/dtslint) is using `"typescript": "next"` which resolves to _2.9.0-dev_ (and npm versions below 3 fail if there's a peerDependency conflict).

I've opened palantir/tslint#3792 which should resolve this errors, but I think it's safe to ignore them for now as they relate to the linter, and not to the actual performance of the library.

@dblock what do you think?

dblock:
Well I don't want to check in something that turns the build to broken. Can we resolve this by locking down things to an older version?

kobim:
_dtslint@0.1.2_ passes the installation, but fails as it uses arrow functions (which are only supported since node v4.)
The current version of _dtslint_ supports `node>=6`, and if we understand that we only use the _dtslint_ at the _Node.js: next_ job of the CI, I think it's safe to remove it from our `devDependencies` (`.travis.yml` already installs it, we just need to make sure it installs the right version.)

I've submitted #329 with this fix attempt, and it seems to address the issue.
If accepted and merged, contributors could sync-merge their PRs and we will be able to address their real commits instead of our build failures :)

CodeCutterUK:
@kobim Got a clean build now, thanks for fixing the `dtslint` issue

dblock:
We can merge this right? @lazerwalker ?

PR #329, opened 2018-04-05
Run dtslint only on the latest node version

kobim:
As dtslint is only required for linting the definition files, and its minimum requirements are `node>=6.0.0` (as of _dtslint@0.2_), there's no need to try and install it for older node versions.

(See issues at alexa-js/alexa-app#324 and palantir/tslint#3792)

# Other options which were rejected
1. Rolling dtslint to v0.1.2 where tslint version is 5.0.0 and maximum TypeScript version is 2.3 (instaed of _next_)
The `npm install` phase passes, but we can't run `npm run dtslint` as it uses arrow functions (which were only introduced at node v4)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ node_js:
   - 5
   - 6
   - 7
+  - 8
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


dblock:
👍

dblock:
Nice work @kobim thanks!

PR #330, opened 2018-04-05
Making coveralls work again

kobim:
This repo's [coveralls page](https://coveralls.io/github/alexa-js/alexa-app) didn't update for a long time (since 9/7/17) due to _istanbul_'s version not being compatible with covering spawned processes. (more on that at [gotwarlost/istanbul#262](https://github.com/gotwarlost/istanbul/issues/262#issuecomment-185239815)).
This PR's coveralls: https://coveralls.io/builds/16371809

Istanbul's 1.x.x-alpha versions solved this issue.
(Istanbul has made [version 2](https://istanbul.js.org/) already, but that's another integration we should plan for later.)

I think this is important as we are about to enter few new features (e.g. #313) and we should make sure  the new code is properly tested and doesn't reduce the coverage of the package.
(We can also grant repo permission to @coveralls and get reports on each PR.)


dblock:
Looks great. @kobim feel free to turn it on for PRs.

PR #331, opened 2018-04-06
Format files more consistently.

dblock:
Without going crazy at introducing formatting tools just run some basic formatting on .js and .json files.


dblock:
I'll self merge this since it's harmless.

PR #344, opened 2018-04-28
Add coveralls as a check for pull requests

kobim:
Following #330, I added coveralls as a status check on pull requests, so we will be able to understand if coverage increased/decreased before we merge (and act accordingly).

In the process, I added/fixed tests to make sure all the branches of the code will get called - reached to 100% coverage across all parameters.


dblock:
👍

PR #352, opened 2018-05-19
allow utterance expansion in custom slot type synonyms

daanzu:


daanzu:
Would need flatMap, not map. Could I interest you in reduce?

kobim:
My bad, I forgot that the synonyms array might expand. I liked the previous implementation better, can you revert to that and I'll merge your changes?

daanzu:
Removed reduce version.

kobim:
Thank you for your contribution!

Issue #355, opened 2018-06-08
Drop node versions 1-4 and add support for node 9 and 10


tejashah88:
Since node 1 to 3 aren't used as much and with the recent announcement of node 4.x reaching end of life, I think it would make sense to not include it for future travis builds. On the other hand, we should add testing for node versions 9 and 10. What do you guys think?

tejashah88:
@alexa-js

dblock:
I think adding new versions is a no brainer. 

I would drop support only if it's in the way, eg. when fixing or adding something and that breaks old versions. If it costs nothing to maintain my vote is to keep it, avoid pain for users who're slower to upgrade.

PR #356, opened 2018-06-09
Add node 9 and 10 to travis builds

tejashah88:
Coming from #355

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,6 +21,8 @@ node_js:
   - 6
   - 7
   - 8
+  - 9
+  - 10
 
 cache:
   directories:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -21,6 +21,8 @@ node_js:
   - 6
   - 7
   - 8
+  - 9
+  - 10
```
kobim: The latest node version (currently 10) is already tested at the `matrix` section (`node_js: node`), so there's no need to add it here (at least until version 11 is released)

tejashah88: Thanks for letting me know!

kobim:
Thank you!

PR #364, opened 2018-08-17
Issue #359 .reprompt() not concatenating

### LINKED ISSUE #359: .reprompt() not concatenating
Current code:
      this.response.response.reprompt.outputSpeech.ssml = SSML.fromStr(str, this.response.response.reprompt.outputSpeech.text);

Correction
      this.response.response.reprompt.outputSpeech.ssml = SSML.fromStr(str, this.response.response.reprompt.outputSpeech.ssml);

Need to reference the .ssml variable (instead of .text)

andrewjhunt:
Fixes #359 
https://github.com/alexa-js/alexa-app/blob/master/index.js#L51

Example: the following will speak "1 2" as output.
  response.reprompt("1").reprompt("2")

Existing unit tests are updated and pass.

NOTE: the doc is not explicit on the expected behavior of multiple reprompts for the same response. I expected that multiple reprompt() would also concatenate as say(). The existing unit tests didn't reflect this but it's not clear that was intentional. Plus the code appeared to reference a non-defined variable.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -199,7 +199,10 @@ export class response {
   /** Empties the response text */
   clear: () => response;
 
-  /** Tells Alexa to re-prompt the user for a response, if it didn't hear anything valid */
+  /**
+   * Tells Alexa to re-prompt the user for a response, if it didn't hear anything valid.
+   * Multiple calls to say() will be appended to each other.
+   */
   reprompt: (phrase: string) => response;
 
   /**
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -201,7 +201,7 @@ export class response {
 
   /**
    * Tells Alexa to re-prompt the user for a response, if it didn't hear anything valid.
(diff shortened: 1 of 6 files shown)


kobim:
Hi @andrewjhunt, can you please update the CHANGELOG and also add a note about the concatenation of `reprompt()`?
Other than that, it looks good to me.

andrewjhunt:
added reprompt() multiple use doc to README.md and index.d.ts

types/index.d.ts
```diff
@@ -199,7 +199,10 @@ export class response {
   /** Empties the response text */
   clear: () => response;
 
-  /** Tells Alexa to re-prompt the user for a response, if it didn't hear anything valid */
+  /**
+   * Tells Alexa to re-prompt the user for a response, if it didn't hear anything valid.
+   * Multiple calls to say() will be appended to each other.
```
kobim: > Multiple calls to **reprompt()**

Issue #358, opened 2018-07-18
Support for including utterances/ samples in slots


abdulkaderjeelani:
Now the schema's slot is an object that takes a map. Doc says "The slots object is a simple name: type mapping. The type must be one of Amazon's built-in slot types, such as AMAZON.DATE or AMAZON.NUMBER."
If we require to specify a list of samples, for a slot it is not possible. This is especially useful when modeling a dialog flow. see https://raw.githubusercontent.com/alexa/alexa-cookbook/master/feature-demos/skill-demo-plan-my-trip/models/en-IN.json

In code it is explicitly passed as empty array. Changing the value of the map from string to object should do.
```

if (intent.slots && Object.keys(intent.slots).length > 0) {
        intentSchema["slots"] = [];
        for (key in intent.slots) {
          //  It's unclear whether `samples` is actually used for slots,
          // but the interaction model will not build without an (empty) array
          intentSchema.slots.push({
            "name": key,
            "type": intent.slots[key],
            "samples": []
          });
        }
      }
      schema.intents.push(intentSchema);
    }
```

dblock:
Care to pull request a fix for this?

matt-kruse:
Comment added on the PR. Merge and release soon?

PR #366, opened 2018-08-21
Intents and slots prompts confirmation

fabien88:
Intent and slots confirmation with samples prompts like seen in  #358

### DIFF
--- a/index.js
+++ b/index.js
@@ -312,6 +312,7 @@ alexa.intent = function(name, schema, handler) {
   this.dialog = (schema && typeof schema.dialog !== "undefined") ? schema.dialog : {};
   this.slots = (schema && typeof schema["slots"] !== "undefined") ? schema["slots"] : null;
   this.utterances = (schema && typeof schema["utterances"] !== "undefined") ? schema["utterances"] : null;
+  this.prompts = schema && typeof schema["prompts"] !== "undefined" ? schema["prompts"] : null;
 
   this.isDelegatedDialog = function() {
     return this.dialog.type === "delegate";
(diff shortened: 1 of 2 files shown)


matt-kruse:
This is a great improvement. I'm trying to add a Dialog to a skill right now and finding that I need the exact changes in this PR. This should be merged along with some other changes and a release made, otherwise this framework is going to fall behind the quickly-evolving API. @dblock any thoughts?

Timvdv:
This is a great feature! Hope it going to be merged soon.

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 * [#344](https://github.com/alexa-js/alexa-app/pull/344): Add coveralls as a check for pull requests - [@kobim](https://github.com/kobim).
 * [#352](https://github.com/alexa-js/alexa-app/pull/352): Allow utterance expansion in custom slot type synonyms - [@daanzu](https://github.com/daanzu).
 * [#361](https://github.com/alexa-js/alexa-app/pull/361): Fix object reference for dialogState in in doc - [@Sephtenen](https://github.com/Sephtenen).
-* Your contribution here.
```
dblock: Please put this back :)

index.js
```diff
@@ -750,16 +769,92 @@ alexa.app = function(name) {
...
+        dialog: {
+          intents: []
+        },
+        prompts: []
+      },
+      intentName,
+      intent,
+      key;
+
+    var creatPrompt = function(promptId, variations) {
```
lazerwalker: This should be named `createPrompt`, yeah?

index.js
```diff
@@ -750,16 +769,92 @@ alexa.app = function(name) {
...
+      return {
+        id: promptId,
+        variations: variations.map(prompt => ({
```
lazerwalker: I wonder if this should check if the user's passed in an already-constructed response object, and if so just use that instead of string-concatenating SSML for them? 

I'm not sure if I'm being silly, and maybe there's no reason anyone would want to do that in practice, but seeing SSML being manually constructed like this usually gives me pause.

lazerwalker:
FYI, the CI failures (at least the ones on the latest node) are coming in via dtslint, the linter for our TypeScript type definitions. You need to add the new properties you've added to the appropriate places in the type definitions.

PR #368, opened 2018-08-29
updated to fix "Error" on ISP requests

User1m:
Fix issue where ISP requests added "'Error: not a valid request" to ssml output

### DIFF
--- a/index.js
+++ b/index.js
@@ -598,6 +598,8 @@ alexa.app = function(name) {
             if (typeof self.sessionEndedFunc == "function") {
               return Promise.resolve(self.sessionEndedFunc(request, response));
             }
+          } else if (requestType && requestType.indexOf('Connections') > -1) {
+            return Promise.resolve(request, response);
           } else if (request.isAudioPlayer()) {
             var event = requestType.slice(12);
             var eventHandlerObject = self.audioPlayerEventHandlers[event];
--- a/index.js
+++ b/index.js
@@ -225,6 +225,10 @@ alexa.request = function(json) {
     }
     return this.data.request.type;
   };
+  this.isInSkillPurchase = function() {
+    var requestType = this.type();
+    return (requestType && 0 === requestType.indexOf("Connections."));
+  };
   this.isAudioPlayer = function() {
     var requestType = this.type();
     return (requestType && 0 === requestType.indexOf("AudioPlayer."));
@@ -598,7 +602,7 @@ alexa.app = function(name) {
             if (typeof self.sessionEndedFunc == "function") {
               return Promise.resolve(self.sessionEndedFunc(request, response));
             }
-          } else if (requestType && requestType.indexOf('Connections') > -1) {
+          } else if (request.isInSkillPurchase()) {
             return Promise.resolve(request, response);
           } else if (request.isAudioPlayer()) {
             var event = requestType.slice(12);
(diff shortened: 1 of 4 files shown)


User1m:
@dblock please review :). 
I think a long term fix for handling when AZ introduces new requests types might be needed but for the time being this solves my problem

index.js
```diff
@@ -598,6 +598,8 @@ alexa.app = function(name) {
             if (typeof self.sessionEndedFunc == "function") {
               return Promise.resolve(self.sessionEndedFunc(request, response));
             }
+          } else if (requestType && requestType.indexOf('Connections') > -1) {
```
dblock: I am going to nitpick a bit. It's unclear what this does, so I would like the method to move into `request` and be similar to `isAudioPlayer` explaining what kind of request this is. Then, it should be checking that it starts with `Connections.` otherwise some future `WhateverConnections` will cause undebuggable behavior. 

Thoughts?

User1m: can do that. can you point me to the line you're referring to with `isAudioPlayer`?

User1m:
Ping @dblock

index.js
```diff
@@ -225,6 +225,10 @@ alexa.request = function(json) {
     }
     return this.data.request.type;
   };
+  this.isInSkillPurchase = function() {
+    var requestType = this.type();
+    return (requestType && 0 === requestType.indexOf("Connections."));
+  };
```
dblock: Can you explain how "Connections." means "isInSkillPurchase"? Point to an Amazon doc?

dblock:
Ok, I read the docs. It's not much :( They don't say anything about these "connections" other than "to implement a purchase request put Connections.Response here" or something like that. 

I understand we're fixing an error for these requests, but it feels to me we want to do more than just resolve the call here. Two thoughts:

- It's only a matter of time till Amazon adds some new type of request/response and maybe our code needs to handle the future better instead of always handing a special case like this one. If that's true, then maybe there's a different/better fix for this specific problem?
- Do we want to handle purchase flow natively? I think the answer is yes. It doesn't mean we don't want a fix for this specific problem, but if we're going to be checking for a specific request type we should be doing more than a `Promise.resolve`, similar to audio requests.

Thoughts? 

Also cc: @lazerwalker who was doing a lot with this and maybe has an opinion

index.js
dblock: If we're going to handle these specifically I think the check should be checking either for "Connections.SendRequest" or "Connections.Response", before we get "Connections.DoSomethingElse" that means ... something else. But see my bigger questions below.

dblock:
And an even more basic question, we have support to handle these via `on`.

```js
app.on('Connections.SendRequest', (request, response, request_json) => {
});
```

So why do we need this code?

matt-kruse:
Yes, these new types of requests will keep coming faster than we can directly support the new features. That's exactly why I added .on() so devs could handle new request types.

Ideally, each new type of functionality that sends new request types will get full support with helpful utilities and built in handling. Eventually.

There is a lot of new stuff coming. Trust me. :)

User1m:
Ok maybe I was just using the sdk wrong. But the issue I was trying to solve was that the SDK kept adding "Error: not a valid request" (INVALID_REQUEST_TYPE) to requests with `Connections.SendRequest`. 

@matt-kruse can you confirm if I move to `app.on('Connections.SendRequest', ...` I will avoid this error?

lazerwalker:
Yep, switching to `app.on` will clear up that exact error; the result should be functionally identical to your proposed patch. If you're curious about the code, I'd poke at https://github.com/alexa-js/alexa-app/blob/master/index.js#L619-L623 and https://github.com/alexa-js/alexa-app/blob/master/index.js#L466-L468. 

Just otherwise wanted to +1 @dblock and @matt-kruse's points here. Eventually, having our own consciously-designed API methods and helpers for the payment flow would rock. But it's probably fairly unstable on Amazon's end, and fairly unclear on our end what the correct abstraction is. `app.on` feels like the right tool for right now.

PR #373, opened 2018-10-14
Keep multiple newlines in card output (Continue PR #83)

fremail:
See PR #83 (2 years old hehe)

I merged the changes to current code and added tests.


kobim:
Hi @fremail, thanks for your contribution!

Can you please sync merge from master so the Travis CI build will pass?

Also, I have the following doubts:
1. If the only call to `cleanse` is in _index.js_, why there's a new parameter to the function? If this is the only usage, it should be the default behavior.
1. I think the reason behind this regex is to remove whitespaces at the end and start of lines, and your code doesn't handle that. Maybe replacing the current patten with a custom version of `\s` will be better? (e.g. `[ \t\r\f]`)
1. Any reason you omitted tests from `test_ssml.js`? As the real change happens in a function related to `to-ssml.js`, I think the tests should occur there.

I think it's time we close this 2 years old issue :)

kobim:
LGTM!
I'll merge it to the master and will it will be released along with 4.2.3 (hopefully along with #378) this weekend.

Thanks again!

Issue #102, opened 2016-12-30
Jump from launch to an initial intent


nikolasleblanc:
If at the conclusion of an intent I would like to return to the initial intent, how would I do that?

```
app.launch(function(request,response) {
  // jump to InitIntent
});
app.intent('InitIntent', function(request, response) {
  // some intent stuff
});
```

dblock:
Maybe ask this on an Alexa forum? (add link)

nikolasleblanc:
It can definitely be done, just wondering if it was part of the alexa-app api, see https://github.com/alexa/alexa-skills-kit-sdk-for-nodejs:

```
var handlers = {
    'LaunchRequest': function () {
        this.emit('HelloWorldIntent');
    },

    'HelloWorldIntent': function () {
        this.emit(':tell', 'Hello World!');
    }
 };
```

dblock:
@nikolasleblanc Would welcome a new feature that does it!

nikolasleblanc:
Cool @dblock. I'm _pretty_ new to this, but really enjoying it. Will see what I can do!

ajcrites:
You can create a function that is used by both the initial intent and the launch intent. I think that it would currently be very difficult to write code that would trigger a handler (launch or otherwise) from another handler.

    app.launch((req, res) => {
      // do launch things
      return initialBehavior(req, res);
    });
    app.intent("HelloWorldIntent", (req, res) => {
      return initialBehavior(req, res);
    });
    function initialBehavior(req, res) {
      // do things you would do for both here
    }

ajcrites:
I like the idea of being able to trigger other handlers from within a handler (launch as well as other intents). I'll try to hack on this when I get a chance to.

kobim:
Resolved via #374

PR #374, opened 2018-10-17
Router

### LINKED ISSUE #262: Way to switch intents?
Is there a way to trigger/switch an intent manually? 

My use case:
An intent is triggered. 
Logic is performed that decides if a user has access to this particular intent. 
If not, then pass the user to another intent on the alexa service and continue the interaction.

fremail:
Hello!

I implemented a router requested in #262.
Actually I brought most code of the router from the incoming request routing based on request type.

In two words, I did two things:

1) Added ability to route request to intent/AudioPlayer event handler/PlaybackController event handler/custom request handler and other request types.
`request.getRouter().launch();`
`request.getRouter().sessionEnded();`
`request.getRouter().intent("oneIntent");`
`request.getRouter().custom("sayHello");`
`request.getRouter().audioPlayer("PlaybackNearlyFinished");`
`request.getRouter().playbackController("NextCommandIssued");`
`request.getRouter().displayElementSelected();`

2) Simplify incoming request routing by using the new router.

resolves #262, #102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 * [#364](https://github.com/alexa-js/alexa-app/pull/364): Fix reprompt() to concatenate multiple SSML prompts - [@andrewjhunt](https://github.com/andrewjhunt).
 * [#371](https://github.com/alexa-js/alexa-app/pull/371): Call response.prepare() after post() - [@fremail](https://github.com/fremail).
 * [#373](https://github.com/alexa-js/alexa-app/pull/373): Keep multiple newlines in card output - [@kielni](https://github.com/kielni) and [@fremail](https://github.com/fremail).
+* [#374](https://github.com/alexa-js/alexa-app/pull/374): Added Router - [@fremail](https://github.com/fremail).
 * Your contribution here.
 
 ### 4.2.2 (April 7, 2018)
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@
     * [slot](#slot)
     * [slotResolution](#slotResolution)
     * [resolutionValue](#resolutionValue)
+    * [router](#router)
 * [Request Handlers](#request-handlers)
     * [LaunchRequest](#launchrequest)
     * [IntentRequest](#intentrequest)
@@ -186,6 +187,9 @@ Boolean request.hasSession()
 // return the session object
 Session request.getSession()
 
+// return the router object
+Router request.getRouter()
+
 // return the request context
 request.context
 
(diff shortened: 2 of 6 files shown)


fremail:
@kobim AFAIK Request will refer to Router (not have a copy of it), so there shouldn't be memory issues.

I did similar to Session object which is in Response. But the Router is in Request because it sounds better: "Route **request** to ...".

There is no problem to separate it from Request if you want.
_________

Also during last several days I'm thinking about some flag for stopping next routing.
For example in `pre` you decide to switch to `myIntent` intent though the original request was to `originalIntent`. And you do not want to execute `originalIntent`. My example in code:
```
// incoming request is to originalIntent
app.pre = function (request, response) {
    return request.getRouter().intent('myIntent').end();
};
app.intent('originalIntent', function () {
    // won't be executed in this example
});
app.intent('myIntent', function (request, response) {
    response.say('Hi there!');
});
```
Though maybe it's worth to be a separate PR.

kobim:
@dblock what do you think?

dblock:
This looks good to me. I say @lazerwalker makes the final call :)

lazerwalker:
Sorry, somehow missed the earlier ping for this! This all LGTM 👍. 

I'm not too concerned about memory leaks — I've never seen circular references like that cause any practical JS issues, and some Googling suggests modern JS engine GCs are generally smart enough to handle situations like that without trouble.

It felt weird at first glance, but thinking about it, I think having router be an attribute on request makes more sense than alternatives. I like how it mirrors how we handle session.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 * [#364](https://github.com/alexa-js/alexa-app/pull/364): Fix reprompt() to concatenate multiple SSML prompts - [@andrewjhunt](https://github.com/andrewjhunt).
 * [#371](https://github.com/alexa-js/alexa-app/pull/371): Call response.prepare() after post() - [@fremail](https://github.com/fremail).
 * [#373](https://github.com/alexa-js/alexa-app/pull/373): Keep multiple newlines in card output - [@kielni](https://github.com/kielni) and [@fremail](https://github.com/fremail).
+* [#374](https://github.com/alexa-js/alexa-app/pull/374): Router - [@fremail](https://github.com/fremail).
```
dblock: Can we change this to "Added Router" or something more detailed?

dblock:
I made a nitpick on changelog, otherwise anyone hit merge pls.

fremail:
Updated Changelog.md.

Squashed and rebased my commits.

PR #376, opened 2018-10-21
Fix node < 5 tests

kobim:
_chai-string_ version 1.5.0 [updated _chai_ dependency to > 4](https://github.com/onechiporenko/chai-string/commit/8fc233c38dedf1f8f733cd290670d7d14dbed3e1#diff-b9cfc7f2cdf78a7f4b91a753d10865a2), hence breaking our devDependencies tree.

Pinning to version 1.4.0 (which was already locked in _package-lock.json_) would resolve this issue (until we will upgrade _chai_ and its dependencies).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "alexa-app",
-  "version": "4.2.2",
+  "version": "4.2.3",
   "lockfileVersion": 1,
   "requires": true,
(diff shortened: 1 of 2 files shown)


kobim:
as there's nothing special about this commit, I'm going to merge it to master so every PR could merge and fix their broken Travis-CI tests.

PR #378, opened 2018-10-22
Implement Alexa CanFulfillIntentRequest 

orlylev:
Add for handler deducted for CanFulfillIntentRequest
Implement default response for Can Fulfill for the intent and slots
Changes to enable working without user id (as needed by AWS)

See details in Amazon documentation:
https://developer.amazon.com/docs/custom-skills/implement-canfulfillintentrequest-for-name-free-interaction.html

https://developer.amazon.com/docs/custom-skills/understand-name-free-interaction-for-custom-skills.html


kobim:
Hi @orlylev, thanks for your contribution!

According to [Alexa's documents](https://developer.amazon.com/docs/custom-skills/understand-name-free-interaction-for-custom-skills.html) it seems like the `CanFulfillIntentRequest` is called along with an intent name and slot, hence a different function should be defined per each intent.
Your implementation defines a single "global function" which receives all the `CanFulfillIntentRequest`s from Alexa. Any reason you chose this implementation? (or am I missing something?)

orlylev:
Hi
I think it should be separate from the regular intent handler because is is
relay not the business logic of implementing the intent.
On the other hand looks like it is a light logic for the intent and list of
slots values that can be common in many cases so I don't fill it should be
a separate handler per intent.
I intent to add basic common logic based on slots list that may work in
most cases.
Orly

types/index.d.ts
```diff
@@ -183,6 +186,9 @@ export class request {
 
   /** @deprecated */
   session: (key: string) => any;
+
+  /** Returns the request CanFulfillIntent */
+  getCanFulfillIntent: () => object;
```
lazerwalker: Can we strongly type this? You should be able to specify that it returns a `canFulfillIntent`.

types/index.d.ts
```diff
@@ -414,3 +426,10 @@ export interface PlaybackController {
   name: string;
   function: RequestHandler;
 }
+
+export class canFulfillIntent {
+  canFulfillIntent: object;
```
lazerwalker: If I understand correctly, this `canFulfillIntent` sub-object is the Amazon type described [here](
https://developer.amazon.com/docs/custom-skills/understand-name-free-interaction-for-custom-skills.html#response-to-canfulfillintentrequest), yeah? 

It'd be great if we could add that custom interface to alexa.d.ts and reference it here. I think making it clear this is the underlying Amazon response object will help clear up the confusion of why a `canFulfillIntent` object itself has a field called `canFulfillIntent`

README.md
```diff
@@ -555,6 +555,28 @@ app.playbackController('NextCommandIssued', (request, response) => {
...
+
+Define the handler for canFulfillIntent used for skill discovery.  For instance the [Implementation instructions](https://developer.amazon.com/docs/custom-skills/implement-canfulfillintentrequest-for-name-free-interaction.html).
+
+See detailed explanation on the Intent and Slot [logic](https://developer.amazon.com/docs/custom-skills/understand-name-free-interaction-for-custom-skills.html).
+
+
+```javascript
+app.displayElementSelected(function(request, response) {
+  // The request object will return CanFulfillIntent object 
+  let canFulfillIntent =  request.getCanFulfillIntent();
```
lazerwalker: I'm not exactly sure what this code is doing — why are you grabbing this variable and then never accessing it?

lazerwalker: I also wonder if it'd be useful to have an example that uses a normal intent instead of a display intent(?). As someone who's only worked on voice skills, I'm having a hard time wrapping my head around how this example relates to existing intent handler code I've written.

lazerwalker:
Hi! Thanks so much for this!

I'm honestly having a hard time making sense of this, even sitting down with both the Amazon documentation and the full patch. I left a few comments along the lines of places where I think some tweaks to documentation might make things more clearer.

It also seems to me like there's a larger philosophical discussion to be had about where canFulfillIntent code should conceptually live. 

Based on the included test files and @kobim's comment earlier, it seems to me like there's a single global function that handles these requests. Is that right? I think I agree with @kobim that having this defined per-intent is more in line with other design decisions this library has made. 

I agree this is separate from the business logic for handling an intent, but I wonder if it should be, say, an additional parameter passed into `app.intent()` when defining a new intent that contains a validation function.

orlylev:
I updated the example (README.md) I hope it is more clear
The flow for canFulfillIntent is difference from real handling the logic of
implementation so I think it should be separate
The response is different and the needed logic is different
It can be break into handler per intent but I feel that global handler is
better in this case. It can be light common logic for many intents.

README.md
lazerwalker: Thanks, the new changes make it much clearer how this works as currently implemented :)

Issue #169, opened 2017-02-10
AudioPlayer.PlaybackFinished can't respond with play directives


wschaeferiii:
Not an issue with the library, rather the readme.

`AudioPlayer.PlaybackFinished` can't respond with play directives, only stop and clear queue directives, per [documentation](https://developer.amazon.com/public/solutions/alexa/alexa-skills-kit/docs/custom-audioplayer-interface-reference#playbackfinished).

On a `app.audioPlayer(PlaybackFinished...` event, which gives throws the error 'Can't respond with more than 0 play directives'.  You'd need to use `app.audioPlayer(PlaybackNearlyFinished...`

dblock:
Would you be so kind to PR a fix? Thanks.

wschaeferiii:
Will do.  Prob this weekend.

kobim:
Resolved via #379

PR #379, opened 2018-10-23
Improve AudioPlayer section in readme

fremail:
Add notes about acceptable responses. See issue #169 

I suppose the change in readme should not appear in changelog since nothing changed in the lib.


kobim:
LGTM
Thank you for your contribution!

PR #381, opened 2018-10-28
Add a test to increase coverage to 100%

kobim:
After #374 merge, the coverage dropped because of a missing branch.
This will restore the 100% coverage.


kobim:
this is just one additional test, so I'm omitting the CHANGELOG modification and merging.

PR #388, opened 2019-05-02
Updating typings for values of AuthorityResolution

camerongraybill:
Typings were messed up, or'd statements were in quotes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### 4.3.0 (Next)
 
-* Your contribution here.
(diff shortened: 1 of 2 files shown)


lazerwalker:
Looks great to me, thanks for this!

Want to bump the changelog and I'll happily merge?

lazerwalker:
Awesome, thanks! :D

camerongraybill:
@lazerwalker Would it be possible to create a patch release with this change? We would like to remove some any-casting we are doing from our code and this is required for that.

PR #393, opened 2019-08-05
Added support for a fallback handler

pariola:
I thought its better to give developers control over the message they want to return when a user's request doesn't match

### DIFF
--- a/index.js
+++ b/index.js
@@ -420,7 +420,8 @@ alexa.router = function(app, request, response, request_json) {
         return Promise.resolve(app.intents[intent].handler(request, response));
       }
     } else {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -413,14 +413,16 @@ alexa.session = function(session) {
 
 alexa.router = function(app, request, response, request_json) {
   this.intent = function(intent) {
+    request.intent = intent;
```
dblock: This modifies the incoming request, I don't think that's a good idea, at the very least it's an anti-pattern to be modifying something coming from the outside.

pariola: Hmm, i understand and already updated the code as that field is still accessible from the raw data

PR #394, opened 2019-08-17
chore: upgrade the entire dev environment

Favna:
This library was in strong need of this, especially examplified by the amount of security vulnerabilities before upgrading:

```sh
npm i
added 404 packages from 865 contributors and audited 2672 packages in 8.639s
found 462 vulnerabilities (150 low, 12 moderate, 299 high, 1 critical)
  run `npm audit fix` to fix them, or `npm audit` for details
```

and after:

```sh
yarn install --audit
yarn install v1.17.3
[1/6] �🔍  Validating package.json...
[2/6] �🔍  Resolving packages...
[3/6] �🔍  Auditing packages...
[4/6] �🚚  Fetching packages...
info fsevents@1.2.9: The platform "win32" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[5/6] �🔗  Linking dependencies...
[6/6] �🔨  Building fresh packages...
0 vulnerabilities found - Packages audited: 885806
✨  Done in 3.70s.
```

So here's what I did in short:
- Upgraded minimum NodeJS to current LTS, v10
- Upgraded all dependencies to their latest possible versions
- Replaced pre-babel-7 packages with their babel-7 equivalents
- Installed additional @types libraries for VSCode / other IDE IntelliSesne
- Rewrote eslintrc to JSON which is the standard format and is easier to modify
- Filtered out package-lock, allowed yarn.lock.
- Configured travis to use yarn for more optimal caching
- Filtered out all log files, not just npm-debug.log
- Renamed all test to the *.spec.js pattern, a common test pattern
- Fixed 3 tests that were failing on  upgraded Chai code
- Linting galore
- Created a "files" field in package.json, so the publish to NPM doesn't have all the unnecessary files.
  - Tarball before: https://hasteb.in/olujuler.sh
  - Tarball after: https://hasteb.in/kumeqewi.sh
- 90f9856cdf31d430047d7e119d4f1fcbd4df1e9d adds slightly better typings
  - Actually utilize the unions that were previously only in comments behind the types. Specifically cardType and Request.
  - Introduce an optional generic for `response.slot()` which now allows structures such as these where the slotName gets limited to the type / enum that is supplied. This is a slight bit more type safety.
image

- Bumped to version 5.0.0. **THIS IS IMPORTANT**. This change entirely includes BREAKING CHANGES and is likely NOT backwards compatible, especially due to the NodeJS bump from v1 to v10. People who still use Node version < 10 can continue to use v4.3.0, but anyone else should be able to upgrade.

Coverage report: https://hasteb.in/docunodu.sh
Sidenote that coverage report didn't work on Windows because Istanbul has some parsing issue with the `_mocha.cmd` file which gets called on it. Works just fine on Linux / MacOS though which should be no problem because Travis is using Ubunutu.

---

@dblock been a while btw hasn't it ;) decided to rewrite my alexa skill to TS as it's one of my last JS based things and I stumbled on all of this.


CHANGELOG.md
```diff
@@ -1,9 +1,14 @@
...
 
-### 4.3.0 (Next)
+### 5.0.0 (Next)
 
+* [#394](https://github.com/alexa-js/alexa-app/pull/394): Upgrade development environment - [@favna](https://github.com/favna).
 * [#388](https://github.com/alexa-js/alexa-app/pull/388): Updating typings for values of AuthorityResolution - [@camerongraybill](https://github.com/camerongraybill).
 
+#### BREAKING CHANGES
+
+* Drop support for all NodeJS versions lower than 10
```
dblock: No need to shout in CHANGELOG :) Put this in https://github.com/alexa-js/alexa-app/blob/master/UPGRADING.md, business as usual.

Favna: I'll move it in a few. I put it in changelog because I personally follow angular's changelog template where this is standard. Didn't know you guys want it in upgrading.

dblock: That's how it was being done until now. We can debate changing that separately I guess, I don't have too strong of an opinion.

Favna: Moved to upgrading in 71875dc616373455b9ad549665a27078c010d5ec

Favna: > That's how it was being done until now. 

Nah it's cool, I personally adopted [commitizen](https://github.com/commitizen/cz-cli) and (autogenerated) [conventional changelog](https://github.com/conventional-changelog/conventional-changelog) in my own projects recently but everyone has their own style

Favna: Concerning this:
> Needs some README changes around compatibility

I changed `4.3.0` to `5.0.0` but I wasn't sure if I should also add a `## Compatibility` block? Not sure to what extend you want to go.

UPGRADING.md
```diff
@@ -1,5 +1,9 @@
 # Upgrading Alexa-app
 
+### Upgrading to >= 5.0.0
+
+* Drop support for all NodeJS versions lower than 10
```
dblock: You should make this a complete sentence. 

_Node.js 10 or later is now required._

dblock:
Quick question: do we *have* to drop node.js support < 10 here? What desn't work?

lazerwalker:
+1, I'm okay with all of this.

Favna:
@dblock 6 or 8 might even be possible but for one I removed bluebird and instead opting for NodeJS's internal promises. But the reason I went for 10 is because anything below is EOL and won't get security updates anymore. 

And it wouldn't exactly be fitting if somehow at some point an exploit would arise in an Alexa skill and it turns out that it's because alexa-app supports insecure legacy code.

dblock:
@Favna See my edits for UPGRADING, I'll merge after that change. Thank you.

Favna:
done @dblock :) !

dblock:
Merged, thank you!

